PR #70, opened 2021-03-25
removed TileItem, import statements switchover WIP

harshadmanglani:
Removed `TileItem`
Started switching all import statements to `package:Medito.....`


.vscode/settings.json
```diff
@@ -0,0 +1,7 @@
+{
```
michaelcspeed, line 1: can you add this to .gitignore? @harshadmanglani

ios/Flutter/flutter_export_environment.sh
```diff
@@ -1,14 +1,13 @@
 #!/bin/sh
```
michaelcspeed, line 1: this too i guess because we'll both be changing it

michaelcspeed (changes requested):

ios/Flutter/flutter_export_environment.sh
harshadmanglani: This one seems to be there under iOS/Xcode related, why isn't it working? Is it because of the additional ** at the beginning of the path?

.vscode/settings.json
harshadmanglani: Done. @michaelcspeed

ios/Flutter/flutter_export_environment.sh
michaelcspeed: I'm not sure.. Where do you see the **? 😕

harshadmanglani: Line 61 of the .gitignore

michaelcspeed: Oh yeah. I'm not sure, gitignore never seems to work properly!
Never mind let's merge anyway. Thanks!

PR #85, opened 2021-04-07
FIX - Downloads were not working on iOS platform

mohit92:
There were two issues mainly:
1. The logic to get path for item was different during downloads and
retrieval
2. The iOS Audio Player requires a file extension to properly play
it. Added mp3 extension while saving the file
3. The toggle for Offline was not deleting the file when selected
No. This was happening due to different file path too


michaelcspeed (approved):
Thank you @mohit92 !

PR #88, opened 2021-04-08
FIX the offline button state inconsistency

mohit92:
1. Check for offline file while opening the session screen so that the
initial state is correct
2. FIX the toggle state when users selected one of Yes/No button

This ensures that the offline toggle is turned On if the download is
already present for this session.


michaelcspeed (approved):

michaelcspeed:
thank you @mohit92 !

PR #89, opened 2021-04-09
Stats row and edit dialog boxes

harshadmanglani:
@michaelcspeed @scouech you might want to take a look at the design for the dialog boxes for editing stats.
The stats row looks like the Figma design - with 'Number of sessions' changed to 'No. of sessions'


michaelcspeed (approved):
thanks @harshadmanglani !

PR #99, opened 2021-04-15
Update UI for Downloads and Favorites Empty screens

mohit92:


lib/utils/strings.dart
```diff
@@ -0,0 +1,3 @@
+const String EMPTY_DOWNLOADS_MESSAGE =
+    'It looks like you haven’t downloaded anything yet. Downloads are useful to save mobile data or to access sessions in places without signal.';
+const String EMPTY_FAVORITES_MESSAGE = 'Favorites are coming soon.';
```
michaelcspeed, line 3: yes! good idea

michaelcspeed (approved):
perfect thanks again @mohit92 !

PR #141, opened 2021-05-11
Added sorting to voicelist in SessionData

### LINKED ISSUE #122: Order narrator/voice
**Screen:** Player option screen

**Affects:** Narrator row

**Currently:** The narrators/voices are returned in a random order.

**Expected behaviour:** We should order narrators from Z to A. 
So if we have Steven, Will, Tiffany, we will get Will, Tiffany, Steven.

jthure:
Added sorting to voicelist. Reverse alphabetical order.
Fixes #122


michaelcspeed:
Hi @jthure 

Thanks for this. I actually implemented this already in the bloc, and it also orders the sessions by length now too 
https://github.com/meditohq/medito-app/blob/6501aa5769ab602b3e6cd6baf6a6c5c5aa07432c/lib/network/session_options/session_options_bloc.dart#L76

Sorry I forgot to close the issue 😞

michaelcspeed:
My solution is kinda ugly though, so any improvements on it are welcome!

PR #144, opened 2021-05-13
test: add some more unit tests for home bloc

qtdzz:
## Description
- Add some small tests for home_bloc
- Add [`equatable`](https://pub.dev/packages/equatable) package (MIT license): this library allows comparing the  equality of the object without writing a lot of boilerplate code such as overriding `==` operator and `hashCode` method. I use this for compare the results in unit tests.
- Add [`mocktail`](https://pub.dev/packages/mocktail) package (MIT license): this library allows mocking objects in a very simple way (you can check the documentation in their package). I need this to mock external dependencies for the testing target. I think it's useful for testing in general. (I tried `mockito` as the flutter documentation but it requires some code generation which is not neat IMHO. So I prefer to use `mocktail` to avoid the code generation... )
- Inject the `HomeRepo` into `HomeBloc` instead of hard-cording in the constructor. This way will help us easily test the `HomeBloc` without caring too much about `HomeRepo`. We can simply mock the HomeRepo and manipulate its data in the way we want.

## Note 
- I noticed that the code base has the coupling issue between layers (Repo-Bloc-Widget) in most of the place. I can start decoupling them one by one when writing tests but it would be nice if we are conscious about this in the future implementation. Please let me know what you think...

## Test
- I manually tested the app. It works :)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -59,7 +59,7 @@ ios/*
 **/ios/Flutter/App.framework
 **/ios/Flutter/Flutter.framework
 **/ios/Flutter/Generated.xcconfig
-ios/Flutter/flutter_export_environment.sh
+**/ios/Flutter/flutter_export_environment.sh
 **/ios/Flutter/app.flx
 **/ios/Flutter/app.zip
 **/ios/Flutter/flutter_assets/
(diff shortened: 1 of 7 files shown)


michaelcspeed (approved):
This looks good, thanks @qtdzz !
Might be a good idea to look into some DI for the repos. Not that they're really complex, but just for cleanliness :)

Issue #127, opened 2021-04-29
Stats background sticks after page loads


michaelcspeed:
**Describe the bug**
When the Home page loads, the stats background stays where it is, while the actual widget itself moves down the page

**To Reproduce**
1. Open the app from cold

**Screenshots**
![image](https://user-images.githubusercontent.com/4272338/116588376-c89ee600-a91b-11eb-8bf7-ee56228c81f6.png)

**Smartphone (please complete the following information):**
 All?

PR #146, opened 2021-05-14
fix: fix sticky stats position on cold start

qtdzz:
## Description
- Fix https://github.com/meditohq/medito-app/issues/127.
- The problem is from the `Ink` widgets which are used to build the `stats_widget` . From their documentation:
```
/// This widget is subject to the same limitations as other ink effects, as
/// described in the documentation for [Material]. Most notably, the position of
/// an [Ink] widget must not change during the lifetime of the [Material] object
/// unless a [LayoutChangedNotification] is dispatched each frame that the
/// position changes. This is done automatically for [ListView] and other
/// scrolling widgets, but is not done for animated transitions such as
/// [SlideTransition].
```
- On the home page, Medito has 3 parts moving at the beginning (which I think we should fix to create better first impression): the announcement animation, shortcuts switching from waiting boxes to the real shortcuts, and courses switching from waiting boxes to the real content. None of these layout changes are notified to the parent widget. Daily messages popping up is also changing the layout but it is a part of ListView, so the `Ink` components response correctly. That's why sometime we see it stuck in different positions. The order of the moving parts affect the stucking position.
- I fixed it by simply wrapping these moving components in a `SizeChangedLayoutNotifier`.  

## Test
- Unfortunately, I can't create any test for this but only manually testing them:

Before: 

https://user-images.githubusercontent.com/11023947/118295643-6548b280-b4dc-11eb-85e4-803b0e506eca.mp4

After:

https://user-images.githubusercontent.com/11023947/118295625-6083fe80-b4dc-11eb-86d0-7c87caccafe5.mp4

### DIFF
--- a/lib/widgets/home/stats_widget.dart
+++ b/lib/widgets/home/stats_widget.dart
@@ -18,15 +18,14 @@ class _StatsWidgetState extends State<StatsWidget> {
       children: [
         Padding(
           padding: const EdgeInsets.only(left: 16.0, bottom: 8.0),
-          child: Text('Stats',
-              style: Theme.of(context).textTheme.headline3),
+          child: Text('Stats', style: Theme.of(context).textTheme.headline3),
         ),
         SizedBox(
           height: 73,
           child: ListView.builder(
             padding: const EdgeInsets.only(left: 8.0, right: 8.0),
             itemBuilder: (context, i) => statsItem(context, i),
-            itemCount: 5,
+            itemCount: 4,
             scrollDirection: Axis.horizontal,
             shrinkWrap: true,
           ),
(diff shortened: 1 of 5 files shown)


lib/widgets/home/stats_widget.dart
```diff
@@ -18,15 +18,14 @@ class _StatsWidgetState extends State<StatsWidget> {
...
-              style: Theme.of(context).textTheme.headline3),
+          child: Text('Stats', style: Theme.of(context).textTheme.headline3),
         ),
         SizedBox(
           height: 73,
           child: ListView.builder(
             padding: const EdgeInsets.only(left: 8.0, right: 8.0),
             itemBuilder: (context, i) => statsItem(context, i),
-            itemCount: 5,
+            itemCount: 4,
```
qtdzz, line 28: I saw of-by-one item here, not sure if it's intentional

michaelcspeed (approved):
Brilliant. Thanks so much @qtdzz !

PR #152, opened 2021-05-23
chore: improve animation on homescreen

qtdzz:
## Problem
I noticed that on cold start, there is a switching moment between loading widgets to the real widgets after getting network data. However, the animation is not smooth and has a stretch circular loading indicator in daily_message_widget. It might not be noticeable on powerful devices with fast network but on slow devices+network (e.g. emulator), it would be very prominent. Here is the comparison before and after my fix:
Before: 
before

After: 
after

## Changes
- Extract `courses_row_item_widget`, `daily_message_item_widget`, `loading_text_box_widget`
- Align the size of loading widgets and real widgets as close as possible
- Add some basic tests for them
- Use AnimationSwitcher to switch between loading and real widget


michaelcspeed (approved):
excellent, thank you @qtdzz !

PR #169, opened 2021-06-19
fix: select current saved background sound in background sheet

qtdzz:
Currently the background is saved and played correctly but it is not reflected when open the background bottom sheet.
This PR fixes it.


michaelcspeed:
Thanks @qtdzz !

Issue #157, opened 2021-05-24
Re-order downloads


scouech:
A few people have asked to have the option to re-order their downloaded sessions. 
We can have a simple drag and drop behaviour to allow re-ordering the list?

I don't think we need to make it obvious...having it as an hidden feature is fine in my opinion. 
If you think it's better to indicate it somewhere, we could use the snackbar we are already using to indicate how to delete a downloaded session and tweak the message to mention the re-ordering..?

mic-ray:
@scouech Maybe a small icon to the right of each downloaded session, which would indicate that the items can be re-ordered, 
 would be an idea? As an example: https://material.io/archive/guidelines/assets/0Bx4BSt6jniD7X0pmWjRNZ21UUFE/components-listcontrols-reorder.png.
 
To comply with your idea of not having it to obvious, maybe these icons only appear once you enter a "list editing mode"? Though i am not sure, whether such a mode is a planned (or even useful) feature.

scouech:
hey @mic-ray , good idea yeah! The "edit mode" thing is a good idea. Maybe a first iteration would be without it, so you can hold and grad to reorder, and then second iteration we could do with an edit mode / icons maybe? Just thinking in terms of making it easy for the dev

mic-ray:
I like the drag and drop idea as a first iteration. Maybe I could try to implement this using the [ReorderableListView](https://api.flutter.dev/flutter/material/ReorderableListView-class.html) widget?

scouech:
sure! Sounds good @mic-ray :)

scouech:
@michaelcspeed if you could review @mic-ray PR would be nice :)

PR #181, opened 2021-06-30
Feature: Reordering of downloaded sessions

mic-ray:
This PR adresses the issue #157.
As discussed in the comments this may serve as a first iteration, where the reordering is done by long pressing on a list item and then dragging the floating item to the wanted position. The [ReorderableListView](https://api.flutter.dev/flutter/material/ReorderableListView-class.html) widget is used to achieve this behaviour.

### The following changes were made:

Downloads widget:
- Refactored the _getDownloadList method to return a ReorderableListView
which provides the onReorder property
- This callback is called once a list item has been reordered and executes
logic to adjust the index of the reordered item
- Furthermore a method to save the new order is called

- Even though a .builder constructor is available for the ReorderableListView
widget, using it caused the Dismissible functionality of the list items to
not work anymore, thus the default constructor is used
- Therefore the itemBuilder prop has been replaced with the children prop
where the download list is mapped to a list of corresponding sliding items

- Added a ValueKey based on the MediaItems id to the outer InkWell of
the Widget returned by the _getSlidingItem method, so that the
ReorderableListView can distinguish between different list items

Downloads Bloc:
- Added a new method to save a given List of MediaItems
- This method is required to save the update order of downloaded sessions

### DIFF
--- a/lib/network/downloads/downloads_bloc.dart
+++ b/lib/network/downloads/downloads_bloc.dart
@@ -81,4 +81,13 @@ class DownloadsBloc {
     list.remove(jsonEncode(mediaFile));
     await prefs.setStringList(savedFilesKey, list);
   }
+
+  // This method is used to save the updated order of the list of downloaded sessions
+  /// Saves a given `List` of `MediaItem` elements (downloaded sessions)
+  static Future<void> saveDownloads(List<MediaItem> mediaList) async {
+    var prefs = await SharedPreferences.getInstance();
+    var list =
+        mediaList.map((MediaItem mediaItem) => jsonEncode(mediaItem)).toList();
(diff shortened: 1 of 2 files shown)


michaelcspeed (approved):
Looks good, thanks @mic-ray

michaelcspeed:
Hey @mic-ray Sorry I had to revert your changes on develop because it was causing some issues with the downloads on the session options screen. The download button no longer animated. Please could you take a look and make a new PR? 
Cheers! :)

mic-ray:
@michaelcspeed Sorry that my changes seemed to cause some issues, I will try and fix them. I'll create a new PR once done :)

Issue #190, opened 2021-08-05
"Pick a narrator & duration" based on context


scouech:
When you open "timer" it says "pick a narrator and duration", but timer does not have a narrator. It should say "pick duration". 
That should be the case for any session with no narrator (none)

mic-ray:
@scouech Do you mind if i take a look at this? 😄

scouech:
Sure @mic-ray  :) Let me know if you need more info!

scouech:
@mic-ray any updates ? :)

mic-ray:
@scouech I am sorry i wasn't able to tackle this earlier. Is it okay with you if i will attempt to solve this (or atleast take a look on it, if i run into problems) during this weekend, or is that too late for you? :)

PR #193, opened 2021-08-19
Fix: Context aware pick narrator/duration text

mic-ray:
This PR adresses the issue #190.

The following changes were made:

**String constants:**
- Adjusted the name of the string for picking a narrator and duration
and adjusted its content to not be upper case (like the other strings)
- Added new string to display when only a duration is has to be picked

**Sessions options screen:**
- Added a boolean to the buildOptionsPanel method which is true, if the
VoiceItems to display contains atleast one narrator, so if the headerValue
on atleast one of the VoiceItems is not empty
- If no narrator is contained in the items, the boolean value is false
- The boolean is used to display the according text string above the
narrator/duration section

_Please note: Some small formatting changes were made (on the sessions options screen), due to different formatter settings_


michaelcspeed:
Great thanks so much @mic-ray !

scouech:
Thanks @mic-ray ! That's a good change!

PR #223, opened 2022-01-09
Fixes #159: Added saving file index

### LINKED ISSUE #159: Remember last selected Voice
When selecting a voice and starting a session, the app should remember your choice and the next time you open a session, if that narrator is available, the length options for that voice should be expanded.

For example in this screen, the last selected voice would have been Morgan:
Screenshot 2021-05-25 at 19 06 57

Currently only the first narrator in the list is expanded
See `lib/network/session_options/session_options_bloc.dart` line 164

DeveloperAlfa:
Based on the conversation on the issue, added the last speaker to sharedPreferences

### DIFF
--- a/lib/network/session_options/session_options_bloc.dart
+++ b/lib/network/session_options/session_options_bloc.dart
@@ -27,11 +27,13 @@ import 'package:Medito/widgets/player/player_widget.dart';
 import 'package:audio_service/audio_service.dart';
 import 'package:flutter/cupertino.dart';
 import 'package:pedantic/pedantic.dart';
+import 'package:shared_preferences/shared_preferences.dart';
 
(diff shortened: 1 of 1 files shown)


michaelcspeed (approved):
nice thank you @DeveloperAlfa !

michaelcspeed:
I had to revert these changes @DeveloperAlfa because there were some issues. Please can you test it and let me know?
Firstly I had a error calling `lastSelectedVoice.isNaN` because `lastSelectedVoice` was null. 
Secondly it seems like you are saving only the index, but not the which session it belongs to. What we need to do is save a pair of values: the session id and the index for that session. 
Feel free to create a new PR. Cheers!

PR #228, opened 2022-01-23
Add speaker data to shared preferences

DeveloperAlfa:
Tested on local environment.

The save code is called at both beginTap and change of selected voice


michaelcspeed (approved):

michaelcspeed:
Thanks! @DeveloperAlfa

PR #313, opened 2023-02-08
iOS Icons fix

michaelcspeed:


KeirLogan:
@michaelcspeed I think your force push overwrote some changes. I think we'll need to revert back to before this, git pull on your local repo, then reapply the changes. Let me know if you'd like to call to discuss this.

Issue #312, opened 2023-02-06
[BUG]  A quick double click on TODAYS'S QUOTE card causes 2 bottom-sheets to appear


mikolajkochanek:
As the title suggests quick double click on daily quote widget shows 2 sheets with sharing options. 
I experienced this problem on LG v30 with android version 9.0. I believe this also applies to IOS. This is a very minor issue, but fixing this would improve UX a little bit.

Bkohler93:
The issue does not occur on a [very simple example](https://gist.github.com/Bkohler93/8175dbc77185fae946427b0a32ae470c) on an iPhone 6s emulator running iOS 16.2. Using the same example it does occur running on Pixel 6 emulator running Android 12. 

A package called [tap_debouncer](https://pub.dev/packages/tap_debouncer) provides a widget that wraps around a `GestureDetector` (or any other widget with an `onTap` callback) and can either prevent taps while the `onTap` method completes or wait a specified duration using a `cooldown` parameter. This solution fixed the issue on the Pixel 6 emulator in the [example above](https://gist.github.com/Bkohler93/d9ec540e5bf5f1db2f1294c8120f8679). An altered (untested) `DailyMessageItemWidget` is written below using this package:

https://gist.github.com/Bkohler93/511062fefd59d5e7160a868ee15b97ee

Alternatively, without the need of another dependency, the `DailyMessageItemWidget` can be converted into a `StatefulWidget` and then use a boolean flag along with `setState` and `Future.delay()` within `_createShareAction()` to set a cooldown on how frequently the quote can be tapped.

scouech:
Thank you for reporting @mikolajkochanek and thanks for the help @Bkohler93.
We are discontinuing the quotes, the next version of the app will not include them. Because of that, I'm closing the ticket. 
Cheers

PR #317, opened 2023-02-14
Double-click on Today's Quote causes two sheets to popup

Bkohler93:
Solves the issue using the `tap_debouncer` package to set a cooldown period between taps when opening the Share widget.


Bkohler93:
closes #312

scouech:
Thanks @Bkohler93 !
Will leave that open for now as we are working on a redesign. Will keep you posted :)

Issue #318, opened 2023-02-19
[Screen] Background sound


scouech:
Figma:
https://www.figma.com/file/E9k4sKjlBDUREveJbUtf4A/Untitled?node-id=573%3A2051&t=uVbJdL1HmNkeSqwC-1

Description:
— You will find on figma (under background sound page) the background sound screen.
— Please check the comments for behavior etc.

Endpoint:
Data should be available when calling the session via /backgroundSounds

PR #323, opened 2023-02-25
Screen background sound #318

osamaasifoak:
Hi @michaelcspeed, please take a look at these items in this PR.

- Background screen repository added
- Background screen view model added
- Background screen added
- Audio Player provider added for background sound
- Ignore the player screen


lib/components/headers/collapsible_header_component.dart
```diff
@@ -112,9 +112,12 @@ class _CollapsibleHeaderComponentState
     return Stack(
       fit: StackFit.expand,
       children: [
-        Image.asset(
-          image,
-          fit: BoxFit.fill,
+        // Image.asset(
```
michaelcspeed: Can we remove this comment?

lib/constants/strings/shared_preference_constants.dart
```diff
@@ -0,0 +1,4 @@
+class SharedPreferenceConstants {
```
michaelcspeed, line 1: nice

lib/routes/routes.dart
```diff
@@ -9,7 +10,8 @@ import 'package:flutter/material.dart';
 import 'package:go_router/go_router.dart';
 
 import '../views/home/home_wrapper_widget.dart';
-import '../views/player/player2/player_view.dart';
+import '../views/player_old/player2/player_view.dart';
+// import '../views/player/player_view.dart';
```
michaelcspeed: Remove comment?

lib/routes/routes.dart
```diff
@@ -34,6 +36,7 @@ final router = GoRouter(
   routes: [
     GoRoute(
         path: HomePath,
+        // builder: (context, state) => BackgroundSoundView(),
```
michaelcspeed: also here

lib/routes/routes.dart
```diff
@@ -174,7 +177,12 @@ MaterialPage<void> getCollectionMaterialPage(GoRouterState state) {
 }
 
 MaterialPage<void> getPlayerMaterialPage(GoRouterState state) {
-  return MaterialPage(key: state.pageKey, child: PlayerWidget());
+  var session = state.extra as SessionModel;
+  return MaterialPage(key: state.pageKey, child: PlayerWidget()
+      // child: PlayerView(
+      //   sessionModel: session,
+      // ),
```
michaelcspeed: also here

lib/views/background_sound/components/custom_track_shape_component.dart
```diff
@@ -0,0 +1,75 @@
+import 'package:Medito/constants/constants.dart';
+import 'package:flutter/material.dart';
+import 'package:google_fonts/google_fonts.dart';
+
+class CustomTrackShapeComponent extends RectangularSliderTrackShape {
```
michaelcspeed: I think we have multiple TrackShapes in the project now, so maybe calling it something like "BackgroundSoundVolumeTrackShape" would be better?

lib/views/player/components/bottom_action_component.dart
```diff
@@ -0,0 +1,66 @@
...
+            () => {},
+          ),
+          width8,
+          _bottomActionLabel(
+            context,
+            'DONWLOAD',
+            () => {},
+          ),
+          width8,
+          // if (_hasBGSound())
```
michaelcspeed: need this?

lib/views/player/components/bottom_action_component.dart
```diff
@@ -0,0 +1,66 @@
...
+          _bottomActionLabel(
+            context,
+            'SOUND',
+            () => {},
+          ),
+          // if (_hasBGSound()) width8,
```
michaelcspeed: need this?

lib/views/player/components/bottom_action_component.dart
```diff
@@ -0,0 +1,66 @@
+import 'package:Medito/constants/constants.dart';
+import 'package:flutter/material.dart';
+
+class BottomActionComponent extends StatelessWidget {
+  const BottomActionComponent({super.key});
+
+  @override
+  Widget build(BuildContext context) {
+    final _speedList = ['X1', 'X1.25', 'X1.5', 'X2', 'X0.6'];
+    var _currentSpeed = 'X1';
```
michaelcspeed, line 9: Actually would be great if we could save the user's chosen speed and auto select that next time they load this screen?

michaelcspeed (commented):
I'm wondering about the use of the word "component" - it seems a bit generic to me. Would be better to use "Widget" or "Shape" or something more descriptive, unless I'm missing something?

Other than that, looks great so far. Thank you!

lib/routes/routes.dart
osamaasifoak: Actually, I was switching between the screens that's why I didn't remove the comment but we don't need it anymore so, I will remove all the comments thanks.

lib/views/background_sound/components/custom_track_shape_component.dart
osamaasifoak: Yes, you're right

lib/views/player/components/bottom_action_component.dart
osamaasifoak: All these changes have been made in the player screen PR.

lib/views/player/components/bottom_action_component.dart
osamaasifoak: removed

michaelcspeed:
What do you think about my comment about the term "component" above? @osamaasifoak 

Is this PR ready to merge? Please ping me on Slack if so because I'm still not getting Github notifications :/

osamaasifoak:
I think you're right. A single component contains multiple widgets and is much more complex than a widget. I will make this change to improve downloading feature PR. Let me know if it works for you

PR #329, opened 2023-03-15
Improve dynamic type handling

michaelcspeed:
Ignored some places where I think dynamic might be necessary and removed dynamic in some places.

Some of the widgets and response classes that still use dynamic will be removed soon anyway - we can use this analyser as a reminder :)


lib/components/webview_component.dart
osamaasifoak, line 1: Nice

lib/services/network/dio_client_provider.dart
osamaasifoak, line 1: I guess connectTimeout doesn't accept Duration because it's an Int property.

michaelcspeed: I'm attempting to update the libs :) it changed to a Duration in [v5.0.0](https://pub.dev/packages/dio/changelog#500)

michaelcspeed:
Woo it works! @osamaasifoak 
At the moment pull requests can be merged even if there are issues (because atm there are a lot) but eventually I want to flip the switch so code cannot be merged unless these checks pass :) So it would be great if you could start looking at the output under “Run dart_code_metrics” and “Analyze” while developing so that we can clear up a lot of these warnings. And at the very least, let's have it so that no new code as any major warnings/issues?
Over time I will also add more rules to DCM (at the moment it's only for use of dynamic) so let's get tidying! 🧹 🪄

Also there is a way to get github to post comments on the PRs when there are issues, but I'll leave that for another day 😓

lib/services/network/dio_client_provider.dart
osamaasifoak: Oh, that makes sense

PR #333, opened 2023-03-22
flutter analyze - handled non-deprecated warnings

Bkohler93:
Handled all warnings that weren't due to a Deprecated tag found in a file. Most of them seemed like just variable name changes would fix them but I didn't think that seemed right. 

The one exception is in `lib/views/player/components/duration_indicator_component.dart` line 14 accessing the stream getter 'audioPositionProvider.stream'.

Wasn't sure how to handle this one.

Most of changes shown are auto-formatting changes.

### DIFF
--- a/lib/network/auth.dart
+++ b/lib/network/auth.dart
@@ -1,8 +1,4 @@
-@Deprecated('Remove after refactor')
 const BASE_URL = '';
-@Deprecated('Remove after refactor')
 const INIT_TOKEN = '';
-@Deprecated('Remove after refactor')
 const CONTENT_TOKEN = '';
-@Deprecated('Remove after refactor')
 const SENTRY_URL = '';


michaelcspeed:
Cool thanks @Bkohler93 ! 
Could you also please check unresolved conflicts?

For the stream, I'm also not sure, maybe @osamaasifoak has an idea.
@osamaasifoak could you also review this once the conflicts have been fixed? Thanks!

osamaasifoak:
Hi @Bkohler93, thanks for mentioning seems like .stream is deprecated now and will be removed in the future version. I will update the code as per the new documentation.

cc: @michaelcspeed 

 @Deprecated(
    '.stream will be removed in 3.0.0. As a replacement, either listen to the '
    'provider itself (AsyncValue) or .future.',
  )

Issue #298, opened 2023-01-23
Improve download feature


michaelcspeed:
https://www.figma.com/file/E9k4sKjlBDUREveJbUtf4A/Untitled?node-id=283%3A701&t=8NTE8WlbQDYPElVl-0

`lib/network/downloads/downloads_bloc.dart`
`lib/audioplayer/download_class.dart`

Currently, if the user wants to download a file they have to go to the Session Options screen, select their desired narrator and length, then press download. If they kill this screen when the download is in progress, the download is cancelled.
We would like it so that the download happens in parallel. We can show a notification for this if necessary.

We should investigate: Can we cache the sound that is played when the screen loads, and then just save that if the user presses the Download button (see figma above)

If we change the way downloads work in the next version, users should still be able to access the downloads from previous versions (Let's deprecate the class, and keep the new code as separate as possible)

michaelcspeed:
Downloaded background sounds should work the same way (but not be shown in the Downloads screen)

michaelcspeed:
DownloadsBlock should also be a riverpod provider

PR #334, opened 2023-03-27
Improve download feature #298

osamaasifoak:
- Downloading feature improved
- Download screen logic changed to show the downloaded session


lib/view_model/background_sounds/background_sounds_viewmodel.dart
```diff
@@ -34,8 +34,13 @@ class BackgroundSoundsNotifier extends ChangeNotifier {
 
   void handleOnChangeSound(BackgroundSoundsModel? _sound) async {
     selectedBgSound = _sound;
-    unawaited(SharedPreferencesService.addStringInSF(
-        SharedPreferenceConstants.bgSound, json.encode(_sound?.toJson())));
+    if (_sound != null) {
+      unawaited(SharedPreferencesService.addStringInSF(
```
michaelcspeed: I just had a moment where I was like "what is SF?" It's better to use clear and descriptive names for variables, functions, etc. This can make the code easier to read and understand for everyone who works with it.
I know I've done that a lot in this codebase before - would be good if we can clear that up as we go :D

lib/views/downloads/downloads_view.dart
```diff
@@ -125,24 +146,37 @@ class _DownloadsListWidgetState extends State<DownloadsListWidget>
...
+        Duration(milliseconds: item.audio.first.files.first.duration)
+            .inMinutes
+            .toString();
+    return PackListItemWidget(
+      PackImageListItemData(
+          title: item.title,
+          subtitle:
+              '${item.audio.first.guideName} — ${_getDuration(audioLength)}',
+          cover: item.coverUrl,
+          // colorPrimary: parseColor(item.extras?[PRIMARY_COLOUR]),
```
michaelcspeed: still needed?

lib/views/packs/pack_list_item.dart
```diff
@@ -79,7 +80,7 @@ class PackListItemWidget extends StatelessWidget {
   Padding _coverImageWidget() {
     return Padding(
       padding: const EdgeInsets.all(8.0),
-      child: data.icon ?? getNetworkImageWidget(data.cover),
+      child: data.icon ?? NetworkImageComponent(url: data.cover!),
```
michaelcspeed, line 83: We should avoid using `!` because if it's null, it'll crash. Instead either check for a null value first, or allow the url to accept a nullable type :)

lib/views/player/components/bottom_actions/components/audio_download_component.dart
```diff
@@ -0,0 +1,90 @@
...
+        child: LinearProgressIndicator(
+          value: _getDownloadProgress(downloadAudioProvider, downloadFileKey),
+        ),
+      ),
+    );
+  }
+
+  double _getDownloadProgress(
+      AudioDownloaderViewModel downloadAudioProvider, String downloadFileKey) {
+    return downloadAudioProvider.downloadingProgress[downloadFileKey]! / 100;
```
michaelcspeed: also here, it it's null we should maybe return 0?

lib/views/player/player_view.dart
```diff
@@ -23,30 +24,51 @@ class PlayerView extends ConsumerStatefulWidget {
...
+        '$sessionId-$fileId${getFileExtension(widget.file.path)}');
+    checkDownloadedFile.then((value) {
+      _audioPlayerNotifier.setSessionAudio(widget.file, filePath: value);
+      _audioPlayerNotifier.currentlyPlayingSession = widget.file;
     });
-    loadBackgroundAudio();
   }
 
-  void loadBackgroundAudio() {
+  void setBackgroundSound(AudioPlayerNotifier _audioPlayerNotifier) {
```
michaelcspeed: Function parameters are already local to the function and cannot be accessed outside of it, so marking them as private with _ is redundant :)

michaelcspeed (changes requested):
Nice! Just a few small comments. Also please take a look at the check annotations on the Files changed tab. And there are some conflicts now - probably just small things

lib/view_model/background_sounds/background_sounds_viewmodel.dart
osamaasifoak: I always try my best to give them a meaningful name. Lol, Idk why I did that but yeah thanks for mentioning it. I will change it.

lib/views/downloads/downloads_view.dart
osamaasifoak: No, I will remove this

lib/views/packs/pack_list_item.dart
osamaasifoak: Hmm, you're right. Let me see what I can do here thanks

lib/views/player/components/bottom_actions/components/audio_download_component.dart
osamaasifoak: I also handled the null case. But I can add a condition here for returning 0.

lib/views/player/player_view.dart
osamaasifoak: I know it seems like I mistakenly did that

lib/views/packs/pack_list_item.dart
osamaasifoak: Can we use a placeholder image if the URL is null?

lib/view_model/background_sounds/background_sounds_viewmodel.dart
michaelcspeed: no worries, we all miss stuff when writing this much code - that's why we do PR reviews :)

lib/views/packs/pack_list_item.dart
michaelcspeed: yeah good idea!

michaelcspeed (approved):

PR #338, opened 2023-03-28
flutter analyze warnings reduced to inhouse Deprecated tags

Bkohler93:
Opened new pull request because last one had unnecessary changes.

Now the only warnings output from `flutter analyze --no-fatal-warnings --no-fatal-infos .` are from in-house Deprecated tags. These tags say things like "data' is deprecated and shouldn't be used. use fields instead" (found in `lib/network/packs/announcement_response.data`), does this mean the only change is to rename the `data` variable to `field` and fix any other issues that come up after doing that?


michaelcspeed:
Thank you for this @Bkohler93 !

PR #349, opened 2023-04-19
Add mini player#331

osamaasifoak:
Things to review in this PR

- The mini-player across the app
- Transition while navigating to the player screen
- Improved audio player logic
- Currently playing session will also be visible on the mobile notification tray.
- View model name changed to providers


lib/network/user/user_utils.dart
```diff
@@ -56,6 +56,7 @@ Future _clearStorage(SharedPreferences prefs) async {
   await prefs.setBool(HAS_OPENED, true);
 }
 
+//ignore: prefer-match-file-name
 class UserRepo {
```
michaelcspeed: maybe this can go in its own file

lib/providers/audio_player/audio_player_provider.dart
```diff
@@ -0,0 +1,180 @@
...
+      updatePosition: sessionAudioPlayer.position,
+      bufferedPosition: sessionAudioPlayer.bufferedPosition,
+      speed: sessionAudioPlayer.speed,
+      queueIndex: event.currentIndex,
+    );
+  }
+
+  bool mediaItemHasBGSound() {
+    print('***********');
+    print(mediaItem.value?.extras?[hasBgSound]);
```
michaelcspeed: can remove these

lib/providers/background_sounds/background_sounds_provider.dart
```diff
@@ -7,10 +7,10 @@ import 'package:Medito/repositories/repositories.dart';
 import 'package:Medito/services/shared_preference/shared_preferences_service.dart';
 import 'package:flutter_riverpod/flutter_riverpod.dart';
 import 'package:riverpod_annotation/riverpod_annotation.dart';
-part 'background_sounds_viewmodel.g.dart';
+part 'background_sounds_provider.g.dart';
 
 @riverpod
-Future<List<BackgroundSoundsModel>> backgroundSounds(BackgroundSoundsRef ref) {
```
michaelcspeed, line 13: what's the rationale behind removing these types? just wondering

lib/providers/session/download_session_provider.dart
```diff
@@ -0,0 +1,66 @@
...
+  );
+  unawaited(ref.refresh(downloadedSessionsProvider.future));
+}
+
+@riverpod
+Future<void> deleteSessionFromPreference(
+  ref, {
+//ignore:avoid-unused-parameters
+  required SessionModel sessionModel,
+  required SessionFilesModel file,
```
michaelcspeed: do we need these if they are unused?

lib/routes/routes.dart
```diff
@@ -120,9 +136,9 @@ GoRoute _getBackgroundSoundRoute() {
   );
 }
 
-GoRoute _getWebviewRoute() {
+GoRoute _getWebviewRoute({bool fromRoot = false}) {
   return GoRoute(
-    path: 'webview',
+    path: fromRoot ? WebviewPath : WebviewPath.replaceFirst('/', ''),
```
michaelcspeed: for these `replaceFirst('/', '')` maybe we could have an ext fun that is super simple called

```
extension SanitisePath on String {
  String sanitisePath() {
    return this.replaceFirst('/', '');
  }
}
```

just makes it a bit neater, what do you think?

lib/views/home/home_wrapper_widget.dart
```diff
@@ -100,4 +104,6 @@ class _HomeWrapperWidgetState extends State<HomeWrapperWidget> {
     });
   }
 
+  // @override
+  // bool get wantKeepAlive => true;
```
michaelcspeed: can we delete this?

lib/routes/routes.dart
```diff
@@ -23,28 +24,43 @@ const String Folder3Path = '/folder/:fid/folder2/:f2id/folder3/:f3id';
 const String Player3 = '/folder/:fid/folder2/:f2id/folder3/:f3id/session/:sid';
 const String UrlPath = '/url';
 const String CollectionPath = '/app';
-const String webviewPath = '/webview';
-const String backgroundSounds = '/backgroundsounds';
+const String WebviewPath = '/webview';
+const String BackgroundSoundsPath = '/backgroundsounds';
```
michaelcspeed: Hmm effective Dart says to prefer lowerCamelCase for consts
https://dart.dev/guides/language/effective-dart/style#prefer-using-lowercamelcase-for-constant-names

What do you think? We would need to update all the others too I suppose, like the ones in Tracking

lib/views/session/session_view.dart
```diff
@@ -30,7 +30,8 @@ class SessionView extends ConsumerWidget {
 
   @override
   Widget build(BuildContext context, WidgetRef ref) {
-    var sessions = ref.watch(sessionsProvider(sessionId: 9));
+    // id=9 for testing
```
michaelcspeed: maybe add a TODO here so we're less likely to miss it?

michaelcspeed (changes requested):

lib/network/user/user_utils.dart
osamaasifoak: For now, I created a user_repository file in the same folder because I think we will change the user repository soon in the authentication flow.

lib/providers/audio_player/audio_player_provider.dart
osamaasifoak: removed

lib/providers/background_sounds/background_sounds_provider.dart
osamaasifoak: It was showing me a warning with a yellow line to remove the type so, that's why I removed the type.

lib/providers/session/download_session_provider.dart
osamaasifoak: We don't need it anymore. I will remove it.

lib/providers/background_sounds/background_sounds_provider.dart
michaelcspeed: Strange I would have thought it would prefer the type! Oh well

lib/components/network_image_component.dart
```diff
@@ -21,6 +27,27 @@ class NetworkImageComponent extends StatelessWidget {
...
+            decoration: BoxDecoration(
+              image: DecorationImage(
+                image: imageProvider,
+                fit: BoxFit.cover,
+              ),
+            ),
+          ),
+          placeholder: (context, url) => _shimmerLoading(),
+          errorWidget: (context, url, error) {
+            print(error);
```
michaelcspeed: can remove this?

lib/providers/background_sounds/background_sounds_provider.dart
osamaasifoak: It's happening in some of the places mostly with the providers.

lib/routes/routes.dart
osamaasifoak: haha, I saw some of them were starting with the camel case so that's I changed it otherwise I always prefer to use the lowerCamelCase for consts. I will change them all.

lib/network/auth.dart
michaelcspeed, line 1: Ah but we need a file with empty values for now, otherwise the build will fail

osamaasifoak: haha, I was testing something lol I will add it back.

michaelcspeed (approved):

michaelcspeed:
@osamaasifoak Approved! Feel free to merge when you're ready

osamaasifoak:
Let's merge it then

PR #356, opened 2023-05-17
User authentication flow

osamaasifoak:
This PR contains

- Authentication flow UI
- Authentication API implementation
- Fix audio player implementation after user authentication flow


lib/constants/strings/string_constants.dart
```diff
@@ -1,48 +1,77 @@
...
-  static const String VOLUME = 'Volume';
-  static const String BEGIN = 'Begin';
-  static const String NONE = 'None';
-  static const String REMOVED = 'Removed';
-  static const String PICK_NARRATOR_AND_DURATION = 'Pick a narrator & duration';
-  static const String PICK_DURATION = 'Pick a duration';
-  static const String SHOW_DOWNLOADS = 'Show downloads';
-  static const String FAVOURITES = 'Favouites';
-  static const String EMPTY_DOWNLOADS_MESSAGE =
+  static const String donwloads = 'Downloads';
```
michaelcspeed: typo!

lib/constants/strings/string_constants.dart
```diff
@@ -1,48 +1,77 @@
...
+  static const String sound = 'Sound';
+  static const String volume = 'Volume';
+  static const String begin = 'Begin';
+  static const String none = 'None';
+  static const String removed = 'Removed';
+  static const String pickNarratorAndDuration = 'Pick a narrator & duration';
+  static const String pickDuration = 'Pick a duration';
+  static const String showDownloads = 'Show downloads';
+  static const String favourites = 'Favouites';
+  static const String emptyDonwloadsMessage =
```
michaelcspeed: typo!

lib/models/folder/folder_model.dart
michaelcspeed, line 1: 'avoid using dynamic type' :D

lib/utils/validation_utils.dart
```diff
@@ -0,0 +1,45 @@
+import 'package:email_validator/email_validator.dart';
+
+const String _invalidEmail = 'Invalid Email.';
+const String _fieldRequired = 'Field is Required';
+const String _invalidInput = 'Invalid Input';
```
michaelcspeed: can these go with the other UI strings for the app? so that when we go to translate them they are all together? or even in a `error_message_constants.dart` file if you think that's better

lib/views/auth/join_intro_view.dart
```diff
@@ -0,0 +1,95 @@
...
+                  Text(
+                    StringConstants.joinTheMeditoFamily,
+                    style: textTheme.headlineMedium?.copyWith(
+                      color: ColorConstants.walterWhite,
+                      fontFamily: ClashDisplay,
+                      height: 2,
+                      fontSize: 24,
+                    ),
+                  ),
+                  for (int i = 0; i < joinBenefitList.length; i++)
```
michaelcspeed: should use brackets here for the for loop

can also do `iterable.forEach((element) { ... });` which is a bit more readable

lib/views/auth/join_verify_OTP_view.dart
```diff
@@ -0,0 +1,227 @@
...
+                fontFamily: DmSans,
+                height: 1.5,
+                fontSize: 14,
+                decoration: auth.counter
+                    ? TextDecoration.none
+                    : TextDecoration.underline,
+              ),
+            ),
+          ),
+          if (auth.counter)
```
michaelcspeed: let's use brackets for all for-loops and if-statements, because for me it's hard to see what is actually in the scope

lib/views/splash_view.dart
```diff
@@ -0,0 +1,74 @@
...
+    });
+  }
+
+  void assignNewTokenToDio(String token) {
+    ref
+        .read(dioClientProvider)
+        .dio
+        .options
+        .headers[HttpHeaders.authorizationHeader] = token;
+  }
```
michaelcspeed: I feel like the view isn't the right place to do this. The UI shouldn't need to know anything about the network layer.
It will also be a lot easier to unit test if we move it to the provider

lib/views/splash_view.dart
```diff
@@ -0,0 +1,74 @@
...
+      initializeAudioPlayer('Bearer ${userTokenModel?.token}');
+      auth.getUserEmailFromSharedPref().then((_) {
+        if (auth.userEmail != null) {
+          context.go(RouteConstants.homePath);
+        } else {
+          context.go(RouteConstants.joinIntroPath);
+        }
+      });
+    }).catchError((e) {
+      initializeUserToken();
```
michaelcspeed: Could this potentially go into an infinite loop? Maybe we need some error message if it fails a few times, or something else, what do you think?

michaelcspeed (changes requested):

lib/utils/validation_utils.dart
osamaasifoak: Yep! You are right. We should move these strings to the string_constants.dart file. Because in the end, they are just a string, it will help us translate all of them simultaneously.

lib/views/auth/join_intro_view.dart
osamaasifoak: We can't use the brackets inside a widget. I extracted the loop as a method.

  Column _benefitPoints(List<String> joinBenefitList, TextTheme textTheme) {
    return Column(
      crossAxisAlignment: CrossAxisAlignment.start,
      children: joinBenefitList
          .map((e) => Text(
                e,
                style: textTheme.labelMedium?.copyWith(
                  color: ColorConstants.walterWhite,
                  fontFamily: ClashDisplay,
                  height: 1.8,
                  fontSize: 16,
                ),
              ))
          .toList(),
    );
  }

lib/views/splash_view.dart
osamaasifoak: Thanks @michaelcspeed for the suggestion. I created a separate provider for this.

lib/views/auth/join_verify_OTP_view.dart
osamaasifoak: Hey @michaelcspeed, we can't use the brackets with if-statements and for for-loops. I extracted the widget as a method now it's more readable. I hope it will help.

` if (auth.counter) _counterAnimation(auth, textTheme),`

lib/components/shimmers/background_sounds_shimmer_component.dart
```diff
@@ -8,14 +8,15 @@ class BackgroundSoundsShimmerComponent extends StatelessWidget {
...
+
     return SingleChildScrollView(
       child: Column(
         children: [
           BoxShimmerComponent(
             height: 130,
           ),
           height8,
-          for (int i = 0; i < 6; i++) _shimmerList(size)
+          for (int i = 0; i < 6; i++) _shimmerList(size),
```
michaelcspeed: Instead of the for-loop you can also do 

`  ...List.generate(6, (_) =>  _shimmerList(size)), `

which might be more readable idk

lib/components/shimmers/session_shimmer_component.dart
```diff
@@ -9,6 +9,8 @@ class SessionShimmerComponent extends StatelessWidget {
   Widget build(BuildContext context) {
     var shimmerList = _shimmerList();
 
+    var list = List.generate(3, (index) => shimmerList).toList();
```
michaelcspeed: Oh yes you've done it here! I like it, can we use it everywhere instead of the forloop?

michaelcspeed (approved):

michaelcspeed:
@osamaasifoak The pipeline is failing because of the analysis, please could you check? Thanks

PR #361, opened 2023-05-29
Renaming sessions to meditations

osamaasifoak:
Sessions/Session name changed to Meditations/Meditation


lib/utils/stats_utils.dart
```diff
@@ -194,28 +194,28 @@ Future<int> _getLongestStreakInt() async {
   return streak ?? 0;
 }
 
-Future<String> getNumSessions() async {
+Future<String> getNumMeditations() async {
   var prefs = await SharedPreferences.getInstance();
 
-  var streak = prefs.getInt('numSessions');
+  var streak = prefs.getInt('numMeditations');
```
michaelcspeed: Be careful here, changing the way stats works will break stats for people already using the app

michaelcspeed (changes requested):

lib/utils/stats_utils.dart
osamaasifoak: You're right. I thought that we are going to change the stats implementation but I was wrong because we also need to send the user stats to the server and this mistake will break the stats for people already using the app. Thanks @michaelcspeed lemme fix this

michaelcspeed (approved):

PR #365, opened 2023-06-03
Internet connectivity

osamaasifoak:


lib/components/errors/connectivity_error_component.dart
```diff
@@ -0,0 +1,64 @@
...
+
+  @override
+  void dispose() {
+    _refreshAPI();
+    super.dispose();
+  }
+
+  void _refreshAPI() {
+    var location = GoRouter.of(context).location;
+    var id = location.split('/');
```
michaelcspeed: The .last() should be here I think right otherwise me get an array?

michaelcspeed (approved):

lib/components/errors/connectivity_error_component.dart
osamaasifoak: ` ref.read(meditationsProvider(meditationId: int.parse(id.last)));`

I was doing something like this. But I think instead of doing this I can get the last element here
`var id = location.split('/').last;`

PR #375, opened 2023-06-13
Bottom sheet UI

osamaasifoak:
This PR contains menu bottom sheet UI and menu items items implementation


lib/views/home/widgets/menu/menu_bottom_sheet_widget.dart
```diff
@@ -0,0 +1,54 @@
...
+                },
+              ),
+            ),
+          ],
+        );
+      },
+    );
+  }
+
+  String getLeadingIconPath(String path) {
```
michaelcspeed: The parameter is more of a name than a path right?

lib/views/home/widgets/menu/row_item_widget.dart
```diff
@@ -0,0 +1,67 @@
...
+    super.key,
+    required this.title,
+    required this.leadingIcon,
+    this.isShowUnderline = true,
+    this.onTap,
+  });
+
+  final String title;
+  final String leadingIcon;
+  final bool isShowUnderline;
```
michaelcspeed: Do you mean _should_ShowUnderline?

lib/views/home/widgets/menu/menu_bottom_sheet_widget.dart
osamaasifoak: You are right thanks

lib/views/home/widgets/menu/row_item_widget.dart
osamaasifoak: Yep!! this variable is only for hiding/showing the underline.

assets/images/ic_edit.svg
michaelcspeed, line 1: Wait so we don't need these anymore do we? We can use

Icon(
  Icons.edit_outlined,
),

Icon(
  Icons.help_outline_outlined,
),

etc?

lib/views/home/widgets/menu/row_item_widget.dart
```diff
@@ -0,0 +1,64 @@
+import 'package:Medito/constants/constants.dart';
+import 'package:flutter/material.dart';
+import 'package:flutter_svg/svg.dart';
+
+class RowItemWidget extends StatelessWidget {
+  const RowItemWidget({
+    super.key,
+    required this.title,
+    required this.iconCodePoint,
+    this.isShowUnderline = true,
```
michaelcspeed: Yeah this should be "shouldShowUnderline". "isShowUnderline" is confusing, 
"isUnderlined" or "hasUnderline" also work

lib/views/home/widgets/menu/row_item_widget.dart
```diff
@@ -0,0 +1,64 @@
...
+  Widget build(BuildContext context) {
+    var border = Border(
+      bottom: isShowUnderline
+          ? BorderSide(
+              width: 0.7,
+              color: ColorConstants.darkMoon,
+            )
+          : BorderSide.none,
+    );
+    var icon = int.parse('0x$iconCodePoint');
```
michaelcspeed, line 29: I would much rather use a mapper class that says like: 

`if(iconStringFromBackend == 'edit') return Icons.edit_outlined` etc. 

@scouech Can you shed some light on this please? Why are we passing the icon code?

lib/widgets/draggable_sheet_widget.dart
```diff
@@ -0,0 +1,32 @@
...
+      expand: false,
+      builder: (
+        BuildContext context,
+        ScrollController scrollController,
+      ) {
+        return Container(
+          decoration: BoxDecoration(
+            color: ColorConstants.onyx,
+            borderRadius: BorderRadius.only(
+              topLeft: Radius.circular(12),
```
michaelcspeed, line 23: Can we make a constants class for these? 
Maybe like MeditoCircularRadiusLarge, MeditoCircularRadiusMedium, MeditoCircularRadiusSmall, something like that?

michaelcspeed (changes requested):

lib/views/home/widgets/menu/row_item_widget.dart
scouech: yeah codepoint :)

lib/constants/strings/asset_constants.dart
scouech, line 1: @osamaasifoak all of this can be removed right?

assets/images/ic_edit.svg
scouech: yep, that should be removed @osamaasifoak

lib/views/home/widgets/menu/row_item_widget.dart
scouech: yes, we decided on using codepoints so that we dont have to translate what the icon name is as a codepoint. We just send a codepoint from the backend. 

However, is there a way @osamaasifoak that if the codepoint is not valid to display a default one? wha would happen if the value is like 0000000 as a codepoint for example?

assets/images/ic_edit.svg
osamaasifoak: Yes @michaelcspeed

lib/views/home/widgets/menu/row_item_widget.dart
osamaasifoak: hmm makes sense

osamaasifoak: Changed to "hasUnderline"

lib/widgets/draggable_sheet_widget.dart
osamaasifoak: Great idea, yes we can do that as well

osamaasifoak: Hey @michaelcspeed, I just checked the problem is we don't have a constant radius in the app. We use a different radius in multiple locations. Let me know if you have any suggestions.

lib/constants/strings/asset_constants.dart
osamaasifoak: Yes, I will remove all the icons that you're giving me in the API

lib/views/home/widgets/menu/row_item_widget.dart
osamaasifoak: Good point @michaelcspeed, Previously, I was showing the icons using the if else statement. But, then we decided to add the codePoints in the API. Because Flutter is also showing the icons using the codePoints.
 e.g. `static const IconData ten_k = IconData(0xe000, fontFamily: 'MaterialIcons');`
 
 Yes @scouech, I will see how to show the default icon.

osamaasifoak: @scouech, just checked we don't have any error callback in IconData or the Icon widget. So the only drawback here is that the PointCodes must be valid. 

@michaelcspeed  do you have any suggestions for us?

michaelcspeed: So what happens if you set an invalid code point? No exception is thrown? Is the widget just empty?

osamaasifoak: No exception and yeah just an empty widget

michaelcspeed: I think that's fine, as long as the app doesn't crash!

lib/widgets/draggable_sheet_widget.dart
michaelcspeed: @scouech Can you check how many there should be please?

michaelcspeed (approved):

PR #376, opened 2023-06-14
Update folder_provider_test.dart

billychili:


michaelcspeed:
@osamaasifoak Could you please take a look at this when you have a sec? It's very small so shouldn't take too long 

To me it looks ok but just want your opinion. Thanks!

osamaasifoak:
@michaelcspeed just went through the changes @billychili, could you tell me why you changed the error test case to the passed test case?

michaelcspeed:
Seems like this PR isn't going anywhere. I will close it. @billychili Feel free to open it again if you get the chance to respond.

PR #379, opened 2023-06-23
Share feature

osamaasifoak:


lib/constants/strings/string_constants.dart
```diff
@@ -77,6 +79,7 @@ class StringConstants {
   static const String buidNumber = 'buidNumber';
 
   //Error
+  static const String someThingWentWrong = '''Something wen't wrong''';
```
michaelcspeed: > wen't 
 :)

lib/utils/utils.dart
```diff
@@ -156,6 +159,31 @@ String getFileExtension(String path) {
...
+    var byteData = await image.toByteData(format: ui.ImageByteFormat.png);
+    var pngBytes = byteData?.buffer.asUint8List();
+
+    // Get the path to the app's temporary directory
+    final directory = await getTemporaryDirectory();
+
+    // Create a file in the temporary directory
+    final file = File('${directory.path}/stats.png');
+
+    // Write the bytes to the file
```
michaelcspeed: We don't need these comments - the code speaks for itself :)

michaelcspeed (changes requested):

michaelcspeed (approved):

PR #382, opened 2023-07-12
Announcement feature

osamaasifoak:


lib/views/home/widgets/announcement/announcement_widget.dart
```diff
@@ -0,0 +1,84 @@
...
+import 'package:Medito/widgets/widgets.dart';
+import 'package:flutter/material.dart';
+
+class AnnouncementWidget extends StatelessWidget {
+  const AnnouncementWidget({super.key, this.announcement});
+  final AnnouncementModel? announcement;
+  @override
+  Widget build(BuildContext context) {
+    return Container(
+      color: ColorConstants.getColorFromString(announcement!.colorBackground),
```
michaelcspeed: If announcement is null, this will crash. Maybe just make it non-nullable (line 9)

michaelcspeed (changes requested):

michaelcspeed:
Looks great, just one small suggestion!

lib/views/home/widgets/announcement/announcement_widget.dart
osamaasifoak: You're right. But, the announcement widget will not show if the announcement is null. So, it won't crash.

michaelcspeed: Yeah but that's a constraint outside this view. In theory I could use this widget somewhere and pass a null announcement and it would crash. And if you wrote a test and passed it a null the test would fail 🤓

osamaasifoak: Lol you are right. The announcement param is nullable Idk why didn't I see the nullable parameter😭

michaelcspeed (approved):

PR #383, opened 2023-07-18
Add event tracking

osamaasifoak:


lib/main.dart
```diff
@@ -99,6 +100,23 @@ class _ParentWidgetState extends ConsumerState<ParentWidget>
 
   @override
   Widget build(BuildContext context) {
+    ref.listen(deviceAndAppInfoProvider, (_, info) {
+      if (info.hasValue) {
+        var val = info.value;
+        var appOpenedModel = AppOpenedModel(
+          deviceOs: val?.os ?? '',
+          deviceLanguage: 'English',
```
michaelcspeed: Can we get this value? The language isn't necessarily English right?

osamaasifoak: Fixed

michaelcspeed (approved):

PR #398, opened 2023-07-25
Add fcm

osamaasifoak:


android/app/google-services-old.json
```diff
@@ -0,0 +1,48 @@
+{
```
michaelcspeed: This file has sensitive information in it. It should not be committed to a public git repo 😬

michaelcspeed (changes requested):

android/app/google-services-old.json
michaelcspeed: I just realised one is "old". I guess that means there was one there already.. can we remove the new one from git?

lib/services/notifications/notifications_service.dart
```diff
@@ -0,0 +1,117 @@
...
+  await initialiazeLocalNotification();
+  var settings = await _messaging.requestPermission(
+    alert: true,
+    badge: true,
+    provisional: false,
+    sound: true,
+  );
+
+  if (settings.authorizationStatus == AuthorizationStatus.authorized) {
+    print('User granted permission');
```
michaelcspeed: Don't need this?

lib/services/notifications/notifications_service.dart
```diff
@@ -0,0 +1,117 @@
...
+    // For handling the received notifications
+    FirebaseMessaging.onMessage.listen((RemoteMessage message) {
+      _showNotification(
+        message.notification?.title,
+        message.notification?.body,
+        json.encode(message.data),
+      );
+    });
+  } else {
+    print('User declined or has not accepted permission');
```
michaelcspeed: I don't think we need this either, we will not see it in prod anyway

lib/services/notifications/notifications_service.dart
```diff
@@ -0,0 +1,117 @@
...
+    iOS: initializationSettingsIOS,
+  );
+  await flutterLocalNotificationsPlugin.initialize(
+    initializationSettings,
+    onDidReceiveNotificationResponse: onSelect,
+  );
+}
+
+void onSelect(NotificationResponse? data) async {
+  print(data);
```
michaelcspeed, line 64: This is a todo I'm guessing?

michaelcspeed (changes requested):

android/app/google-services-old.json
osamaasifoak: Sure, I can remove it no problem. I noticed it was already present in the develop branch and I initially thought it was uploaded by someone else. However, I will go ahead and remove the new one. ;)

lib/services/notifications/notifications_service.dart
osamaasifoak: Haha yes, I will create another PR for this

michaelcspeed (approved):

PR #419, opened 2023-08-12
Allow notification screen

osamaasifoak:


lib/services/network/dio_client_provider.dart
```diff
@@ -33,8 +33,7 @@ final dioClientProvider = Provider<DioApiService>((ref) {
   ));
   dio.interceptors.add(InterceptorsWrapper(
     onError: (e, handler) {
-      // ref.read(authProvider);
-      onError(e, handler, dio, ref);
+      // onError(e, handler, dio, ref);
```
michaelcspeed: @osamaasifoak What should happen here?

michaelcspeed (approved):
I'll approve this so you can merge it when it's ready, but please check that other comment first :) @osamaasifoak

lib/services/network/dio_client_provider.dart
osamaasifoak: Removing this for now

PR #429, opened 2023-08-22
Fix notification destination

osamaasifoak:
* Notification on tap event handled
* Now using Riverpod for Go router navigation


lib/services/notifications/notifications_service.dart
```diff
@@ -30,32 +33,37 @@ Future<AuthorizationStatus> checkNotificationPermission() async {
...
+  FirebaseMessaging.onMessage.listen((RemoteMessage message) {
+    _showNotification(
+      message.notification?.title,
+      message.notification?.body,
+      json.encode(message.data),
+    );
+  });
+}
+
+Future<void> initialiazeLocalNotification(WidgetRef ref) async {
```
michaelcspeed: Spelling!

michaelcspeed (approved):

PR #435, opened 2023-08-28
Fix background audio event

osamaasifoak:


lib/main.dart
```diff
@@ -26,14 +25,14 @@ import 'package:flutter/services.dart';
...
+import 'package:just_audio/just_audio.dart';
 import 'package:shared_preferences/shared_preferences.dart';
 import 'package:Medito/providers/providers.dart';
-
 import 'services/notifications/notifications_service.dart';
 
 late SharedPreferences sharedPreferences;
 late AudioPlayerNotifier audioHandler;
-
+const audioBgEvent = 'com.medito.audio.bg.event';
```
michaelcspeed: Why is this in main.dart?

lib/main.dart
```diff
@@ -108,35 +116,54 @@ class _ParentWidgetState extends ConsumerState<ParentWidget>
...
 
     return MaterialApp.router(
       routerConfig: goRouter,
       theme: appTheme(context),
       title: ParentWidget._title,
     );
   }
+  
+  void _handleAudioCompletion(
+    WidgetRef ref,
```
michaelcspeed: I don't think this should be in main either. Just pass the extras. That will make it easier to test

lib/main.dart
```diff
@@ -108,35 +116,54 @@ class _ParentWidgetState extends ConsumerState<ParentWidget>
...
+      audioProvider.pause();
+
+      WidgetsBinding.instance.addPostFrameCallback((_) {
+        ref.read(audioPlayPauseStateProvider.notifier).state =
+            PLAY_PAUSE_AUDIO.PAUSE;
+      });
+    }
+  }
+
+  void _handleAudioCompletionEvent(
```
michaelcspeed: This too

lib/main.dart
osamaasifoak: Sorry, I have tried many different solutions to fix this issue. I wrote this line for a background service but we don't need it anymore I'm removing it now thanks @michaelcspeed

PR #447, opened 2023-09-05
Fix bg audio

osamaasifoak:


michaelcspeed:
@osamaasifoak there are a few PRs here that were merged before anyone reviewed them... Any particular reason for that?

PR #448, opened 2023-09-07
Selectable text

osamaasifoak:
This PR contains
* Selectable text
* Fixed blur image issue while sharing stats


michaelcspeed:
It's good practice to put separate features in separate PRs. It makes the review process much easier, and if there is an issue in the future, it's easier to trace where it comes from since each PR gets squashed into one commit. So please let's try and do that in the future :)

PR #453, opened 2023-09-14
Remove google fonts

osamaasifoak:


lib/utils/utils.dart
```diff
@@ -144,11 +144,12 @@ String getFileExtension(String path) {
   return '.${path.substring(path.lastIndexOf('.') + 1)}';
 }
 
-Future<File?> capturePng(GlobalKey globalKey) async {
+Future<File?> capturePng(BuildContext context, GlobalKey globalKey) async {
```
michaelcspeed, line 147: What's the context needed for here?

osamaasifoak: I used it to get the device pixel ratio for the image.
`var pixelRatio = MediaQuery.of(context).devicePixelRatio;`

michaelcspeed: Oh yeah, I see. I couldn't see on mobile for some reason!

As a general rule it's best to pass the pixel ratio itself as a parameter. Because when you unit test these methods you would have to mock the MediaQuery/content which might be difficult. But this method has a lot of dependencies anyway so might my impossible to test, I dunno.

michaelcspeed (approved):

PR #456, opened 2023-09-15
Join screen trigger

osamaasifoak:


lib/routes/routes.dart
```diff
@@ -38,39 +38,55 @@ final router = GoRouter(
...
-      pageBuilder: (context, state) => MaterialPage(
-        key: state.pageKey,
-        child: JoinIntroView(),
-      ),
+      pageBuilder: (context, state) {
+        final screen = state.extra != null ? state.extra as Map : null;
+
+        return MaterialPage(
+          key: state.pageKey,
+          child: JoinIntroView(fromScreen: screen?['screen'] ?? Screen.splash),
```
michaelcspeed: Should 'screen' be a constant here?

lib/routes/routes.dart
```diff
@@ -38,39 +38,55 @@ final router = GoRouter(
...
       path: RouteConstants.joinVerifyOTPPath,
       pageBuilder: (context, state) {
-        final data = state.extra! as Map;
+        final data = state.extra != null ? state.extra as Map : null;
 
         return MaterialPage(
           key: state.pageKey,
-          child: JoinVerifyOTPView(email: data['email']!),
+          child: JoinVerifyOTPView(
+            email: data?['email'] ?? '',
```
michaelcspeed: Same for these

PR #465, opened 2023-09-25
Handle user session

osamaasifoak:


lib/views/splash_view.dart
```diff
@@ -1,4 +1,8 @@
+import 'dart:io';
```
michaelcspeed: there's a lot of logic here that shouldn't be in the view. Remember it's good to keep the views as dumb as possible

michaelcspeed (changes requested):

lib/views/splash_view.dart
osamaasifoak: Alright, I'll take care of it.

michaelcspeed (approved):

Issue #431, opened 2023-08-23
Animate row items


scouech:
Could we animate the row items on tap? 
[Here is a link to Slack](https://meditofoundation.slack.com/archives/C04LD4ZQ91N/p1692819886724819) of a recording showing what I mean.
The animation "shrinks" the item on tap, leading to its destination. 

Cheers!

osamaasifoak:
Added!!

scouech:
It should scale the entire view, including the text, not just the background. Because the text otherwise is moving and the items in the row too (the items should not move, it should just shrink the tapped item.)

scouech:
@osamaasifoak could we remove it? it still shows on the testflight build. You can also remove the related library. Cheers

PR #495, opened 2023-10-05
Animate row items#431

osamaasifoak:
[Animate row items#431](https://github.com/meditohq/medito-app/issues/431)


michaelcspeed (approved):
This one only has 1 file change in it and it's Podfile.lock😕

Issue #458, opened 2023-09-17
transferStats event not working properly


scouech:
There seems to be a problem with sending the listened-to data. If the list of listened-to sessions is long, sometimes what we receive on the server is truncated, or not received at all.
I tried from postman with a very long list, and it is posted in the database correctly. But if I go to the old version of the app, listen to a lot of sessions, then update to the new version, the whole list is not sent.
I’ve tried on postman with a list of ids of more than 100 items and it works fine, even multiple triggered at the same time.

scouech:
Related to [#342](https://github.com/meditohq/medito-app/issues/342)

PR #501, opened 2023-10-06
Fix transfer stats #458

osamaasifoak:
[transferStats event not working properly#458](https://github.com/meditohq/medito-app/issues/458)


lib/repositories/stats/stats_repository.dart
```diff
@@ -34,6 +36,7 @@ class StatsRepositoryImpl extends StatsRepository {
   Future<Map<String, dynamic>?> fetchStatsFromPreference() async {
     try {
       var pref = await SharedPreferences.getInstance();
+      // await addDummyStats();
```
michaelcspeed: Remove comment

lib/repositories/stats/stats_repository.dart
```diff
@@ -82,10 +89,174 @@ class StatsRepositoryImpl extends StatsRepository {
...
+      25,
+      26,
+      27,
+      28,
+      29,
+      30,
+      31,
+      32,
+      33,
+      34,
```
michaelcspeed: What is all this? Do we need this in develop?

osamaasifoak: I added this code for testing long stats transfer. Let me remove it now

Issue #468, opened 2023-09-26
Audio player: Restart proxy in case of -1004 error


scouech:

scouech:
Will close this, will keep an eye on sentry see if this changes anything. Thanks @osamaasifoak

PR #510, opened 2023-10-06
Fix audio player proxy

osamaasifoak:
[Audio player: Restart proxy in case of -1004 error#468](https://github.com/meditohq/medito-app/issues/468)


pubspec.yaml
```diff
@@ -27,7 +27,12 @@ dependencies:
   flutter_markdown: ^0.6.7
   shared_preferences: ^2.0.5
   auto_size_text: ^3.0.0-nullsafety.0
-  just_audio: ^0.9.34
+  # just_audio: ^0.9.34
+  just_audio:
+    git: 
+      url: https://github.com/rwrz/just_audio.git
+      ref: fix_cache
+      path: just_audio
```
michaelcspeed, line 35: This is kind of dangerous because when just_audio gets updated, we wont get those changes, since we're now using a different fork. I guess we'll have to find a solution to that at some point

michaelcspeed (approved):

Issue #472, opened 2023-09-28
When cold start, open downloads screen if no internet connection


scouech:
1. Make sure the app is killed
2. Turn on airplane mode (and deactivate wifi)
3. Open the app
The app gets stuck on the splash screen and an error is shown:
"Error occurred while Communication with Server"

The behaviour should be that the app opens directly into the downloads screen, allowing people to listen to downloaded tracks.

PR #512, opened 2023-10-06
Fix open download screen cold start

osamaasifoak:
[When cold start, open downloads screen if no internet connection#472](https://github.com/meditohq/medito-app/issues/472)


lib/views/packs/pack_list_item.dart
```diff
@@ -82,7 +83,11 @@ class PackListItemWidget extends StatelessWidget {
   Padding _coverImageWidget() {
     return Padding(
       padding: const EdgeInsets.all(8.0),
-      child: data.icon ?? NetworkImageWidget(url: data.cover!),
+      child: data.icon ??
+          NetworkImageWidget(
+            url: data.cover!,
```
michaelcspeed, line 88: Don't use the `!` , because it will crash the app if it's null. Better to check if it's null and use a default image if it is.

michaelcspeed (approved):

lib/views/packs/pack_list_item.dart
osamaasifoak: Alright let me fix this

Issue #522, opened 2023-10-09
Change audio library


scouech:
Ok so the error we are getting with the current audio player doesn't seem to be fixed despite following the fix given by the github community. 
When it happens we get error `-1004` and it makes the iOS app unusable. 
With Mike we are thinking it would be worth trying a different library. 
Could you branch of develop and replace `just_audio` with the following library:

Pub.dev: [https://pub.dev/packages/audioplayers](https://pub.dev/packages/audioplayers)

Github: [https://github.com/bluefireteam/audioplayers](https://github.com/bluefireteam/audioplayers)

We should keep the `audio_service` package. 

Could you please give an estimate of how long it would take to swap the library after reading the documents below? Thank you!

You can read about the integration about the two libraries here:
[https://denis-korovitskii.medium.com/flutter-demo-audioplayers-on-background-via-audio-service-c95d65c90ae1](https://denis-korovitskii.medium.com/flutter-demo-audioplayers-on-background-via-audio-service-c95d65c90ae1)

You can read the plugin documentation here:
[https://github.com/bluefireteam/audioplayers/blob/main/getting_started.md](https://github.com/bluefireteam/audioplayers/blob/main/getting_started.md
)

The troubleshooting guide here:
[https://github.com/bluefireteam/audioplayers/blob/main/troubleshooting.md](https://github.com/bluefireteam/audioplayers/blob/main/troubleshooting.md)

scouech:
Instead of changing the library, @osamaasifoak is trying something different with the cache mechanism

PR #525, opened 2023-10-12
Fix audio player

osamaasifoak:
[Change audio library#522](https://github.com/meditohq/medito-app/issues/522)

Initially, we had planned to switch to a different audio library, but for some reason, we decided to limit ourselves to the current one. This pull request includes the implementation of audio caching to fix the audio player problem.

### DIFF
--- a/lib/providers/player/audio_player_provider.dart
+++ b/lib/providers/player/audio_player_provider.dart
@@ -41,8 +41,11 @@ class AudioPlayerNotifier extends BaseAudioHandler
       } else {
         pauseBackgroundSound();
       }
-    } catch (e) {
-      print(e.toString());
+    } catch (err) {
+      unawaited(Sentry.captureException(
+        err,
+        stackTrace: err,
+      ));
     }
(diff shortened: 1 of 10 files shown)


lib/providers/player/audio_player_provider.dart
```diff
@@ -33,13 +33,17 @@ class AudioPlayerNotifier extends BaseAudioHandler
...
+    try {
+      unawaited(trackAudioPlayer.play());
+      var checkBgAudio = mediaItemHasBGSound();
+      if (checkBgAudio) {
+        playBackgroundSound();
+      } else {
+        pauseBackgroundSound();
+      }
+    } catch (e) {
+      print(e.toString());
```
michaelcspeed: This isn't really helpful in prod, should we do a sentry call here?

osamaasifoak: Oh yeah sure

michaelcspeed (approved):

Issue #487, opened 2023-10-05
Fix Retry Mechanism


scouech:
When the an error is thrown during initialisation, the app tries to retry but an error occurs.

To reproduce this, throw an error in `appInitializationProvider` (e.g. alter the INIT_TOKEN and rerun the app)

` Unhandled Exception: 'package:riverpod/src/framework/element.dart': Failed assertion: line 476 pos 11: '_debugCurrentlyBuildingElement == null ||
              _debugCurrentlyBuildingElement == this': Providers are not allowed to modify other providers during their initialization.
`

Please also write unit tests to test these scenarios.

scouech:
@osamaasifoak that has not been done I think?

osamaasifoak:
Hey @scouech, the task has been completed thanks.

PR #526, opened 2023-10-13
Fix retry mechanism#487

osamaasifoak:
[Fix Retry Mechanism#487](https://github.com/meditohq/medito-app/issues/487)

### DIFF
--- a/lib/providers/auth/app_initialization_provider.dart
+++ b/lib/providers/auth/app_initialization_provider.dart
@@ -4,7 +4,6 @@ import 'package:Medito/constants/constants.dart';
 import 'package:Medito/models/models.dart';
 import 'package:Medito/providers/providers.dart';
 import 'package:Medito/services/network/dio_client_provider.dart';
-import 'package:Medito/widgets/widgets.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter_riverpod/flutter_riverpod.dart';
 import 'package:go_router/go_router.dart';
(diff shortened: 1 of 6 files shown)


lib/providers/auth/app_initialization_provider.dart
```diff
@@ -13,31 +13,59 @@ final retryCounterProvider = StateProvider<int>((ref) => 0);
...
   } else {
-    context.go(RouteConstants.downloadsPath);
+    if (shouldNavigate) {
+      ref.read(retryCounterProvider.notifier).update((state) => 0);
+      context.go(RouteConstants.downloadsPath);
+    } else {
+      _recallProvider(ref, context);
+      if (retryCount < 2) {
+        showSnackBar(context, StringConstants.timeout);
+      }
```
michaelcspeed, line 54: This will never be reached, will it? You're saying 

```
if (retry < 2 ) {
} else {
 if (retry < 2) { 
 }
}
```

michaelcspeed: I also noticed in auth_provider generateUserToken, notifyListeners is being called twice. @osamaasifoak

lib/providers/auth/app_initialization_provider.dart
```diff
@@ -13,31 +13,59 @@ final retryCounterProvider = StateProvider<int>((ref) => 0);
...
+    if (retryCount < 2) {
+      _checkInternetConnectivity(context, ref);
+    }
     await _authProvider.initializeUser();
-    var _user = _authProvider.userRes.body as UserTokenModel;
-    _assignNewTokenToDioAndAudioPlayer(ref, _user.token);
-    var deviceInfo = await ref.read(deviceAndAppInfoProvider.future);
-    var data = _setAppOpenedModelData(deviceInfo);
+    var res = _authProvider.userRes;
+    if (res.body != null) {
```
michaelcspeed, line 24: When you are setting the userRes to loading in generateUserToken(), the body is null, so won't mean that handleRetry is called here?

`generateUserToken` and `sendOTP` both call `notifyListeners` twice. @osamaasifoak  Is that deliberate?

michaelcspeed:
@osamaasifoak  I update the app earlier to the new version and it gave me a brand new user. I think there much be an error here somewhere.

appInitializationProvider is confusing, and it might be doing too much.  
It is:
- checking the internet connection
- initialising the user
- assigning the the token to dio and the audio player
- handling the device info
- something with tracking
- retry mechanism

Writing unit tests would ensure that the logic here is sound. This is a crucial part of the app so it would make sense to get some coverage of it.

PR #547, opened 2023-10-19
fix: back button fixed

### LINKED ISSUE #539: Problem with navigation on background sounds (Android)
See Discord report here:
[https://discord.com/channels/1137078055358173226/1162522266987409478](https://discord.com/channels/1137078055358173226/1162522266987409478)

osamaasifoak:
https://github.com/meditohq/medito-app/issues/539

### DIFF
--- a/lib/views/player/player_view.dart
+++ b/lib/views/player/player_view.dart
@@ -86,13 +86,14 @@ class _PlayerViewState extends ConsumerState<PlayerView>
   }
 
   Future<bool> _onWillPop(BuildContext context) async {
-    var currentPage = ref.read(pageviewNotifierProvider).currentPage;
+    var currentPageIndex = ref.read(pageviewNotifierProvider).currentPage;
+    var playerPageIndex = 1;
     var currentLocation = GoRouter.of(context).location;
     if (currentLocation.contains(RouteConstants.backgroundSoundsPath)) {
       context.pop();
 
       return true;
-    } else if (currentPage == 1) {
+    } else if (currentPageIndex == playerPageIndex) {
       ref.read(pageviewNotifierProvider).gotoPreviousPage();
 
       return true;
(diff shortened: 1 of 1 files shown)


lib/views/player/player_view.dart
```diff
@@ -84,8 +85,14 @@ class _PlayerViewState extends ConsumerState<PlayerView>
...
-  Future<bool> _onWillPop() async {
-    if (ref.read(pageviewNotifierProvider).currentPage == 1) {
+  Future<bool> _onWillPop(BuildContext context) async {
+    var currentPage = ref.read(pageviewNotifierProvider).currentPage;
+    var currentLocation = GoRouter.of(context).location;
+    if (currentLocation.contains(RouteConstants.backgroundSoundsPath)) {
+      context.pop();
+
+      return true;
+    } else if (currentPage == 1) {
```
michaelcspeed: I don't understand this. What is 1? Should 1 be in a constant maybe? Seems like a hack.. is there a better way to do it?

osamaasifoak: When we were on the player screen and pressed the back button, the route would be popped from the stack. However, in reality, pressing the back button on the player screen should scroll down and open the page above, such as the home screen, packs or tracks. To address this, I am checking if the current page is equal to 1. If it is, then I take it back to the previous page in PageView because PageView always contains 2 pages. A value of 0 indicates that you are on the home/track/packs screen, and a value of 1 indicates that you are on the player screen.

michaelcspeed: Ahh ok so it's the index of the PageView, sorry I missed that (the formatting is on the phone browser). Could we name it currentPageIndex or something? Might make it clearer. Thanks!

michaelcspeed: @osamaasifoak I will approva this PR. Feel free to merge once you've made the change. Cheers!

michaelcspeed (approved):

lib/views/player/player_view.dart
osamaasifoak: Fixed!!

PR #555, opened 2023-10-24
chore: code refactored

### LINKED ISSUE #551: Add audioCompleted + audioCompleted event to downloads
Could we add the event audioCompleted (with updateStats = true) to downloaded tracks? 
I don't know if you have the audioFileId and trackId for downloads? 
In a similar way, can we add audioStarted to downloads too? Again, I don't know if you have all the data required? Let me know! Cheers

osamaasifoak:
[Add audioCompleted + audioCompleted event to downloads#551](https://github.com/meditohq/medito-app/issues/551)

### DIFF
--- a/lib/views/downloads/downloads_view.dart
+++ b/lib/views/downloads/downloads_view.dart
@@ -76,7 +76,7 @@ class _DownloadsViewState extends ConsumerState<DownloadsView>
           );
         });
       },
-      children: tracks.map((item) => _getSlidingItem(item, context)).toList(),
+      children: tracks.map((item) => _getSlidingItem(item)).toList(),
     );
   }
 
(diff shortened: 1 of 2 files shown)


lib/views/downloads/downloads_view.dart
```diff
@@ -172,4 +158,21 @@ class _DownloadsViewState extends ConsumerState<DownloadsView>
...
+  void _handleDismissable(DismissDirection _, TrackModel item) {
+    if (mounted) {
+      var firstItem = item.audio.first.files.first;
+      ref.watch(audioDownloaderProvider).deleteTrackAudio(
+            '${item.id}-${firstItem.id}${getAudioFileExtension(firstItem.path)}',
+          );
+      ref.read(deleteTrackFromPreferenceProvider(
+        file: firstItem,
+      ).future);
+    }
```
michaelcspeed: This logic shouldn't be in the view should it? It would be better to put it in a provider, then just have the snackbar in the view

michaelcspeed:
I like this refactoring initiative though :D

michaelcspeed (approved):

PR #562, opened 2023-10-28
Search initial page

### LINKED ISSUE #558: Search result initial page
When tapping on "Search" from the home screen, we want to showcase all our packs.
In order to do that we have a new list layout using cards. 

You can see the design here:
[https://www.figma.com/file/E9k4sKjlBDUREveJbUtf4A/Medito-2023?type=design&node-id=1990%3A12280&mode=design&t=v2odjT4iKQrDEyU4-1](https://www.figma.com/file/E9k4sKjlBDUREveJbUtf4A/Medito-2023?type=design&node-id=1990%3A12280&mode=design&t=v2odjT4iKQrDEyU4-1)

- [x]  The interaction with the search bar does not change
- [x]  If you scroll the keyboard should hide like the normal search results
- [x] Each card should have a flexible layout (not a fixed width)
- [x] The cover image might not be present (null or field not returned)
- [x] The subtitle might not be present (null or field not returned)
- [x] If you tap on the card it should go to its path (using type / path, can be a track, pack, link, url) 
- [x] We should be able to re-use the card layout elsewhere in the app ⚠ 

I will provide the endpoint probably tomorrow :) 

Let me know if something is unclear.

osamaasifoak:
[Search result initial page#558](https://github.com/meditohq/medito-app/issues/558)

### DIFF
--- a/lib/views/search/search_view.dart
+++ b/lib/views/search/search_view.dart
@@ -1,13 +1,17 @@
+import 'package:Medito/providers/providers.dart';
 import 'package:flutter/material.dart';
+import 'package:flutter_riverpod/flutter_riverpod.dart';
 import 'widgets/search_appbar_widget.dart';
 import 'widgets/search_initial_page_widget.dart';
 import 'widgets/search_result_widget.dart';
 
-class SearchView extends StatelessWidget {
+class SearchView extends ConsumerWidget {
   const SearchView({super.key});
 
   @override
-  Widget build(BuildContext context) {
+  Widget build(BuildContext context, WidgetRef ref) {
+    var searchQuery = ref.watch(searchQueryProvider);
+
     return Scaffold(
       appBar: PreferredSize(
         preferredSize: Size.fromHeight(kToolbarHeight),
@@ -16,10 +20,14 @@ class SearchView extends StatelessWidget {
       resizeToAvoidBottomInset: false,
       body: Column(
         children: [
-          SearchInitialPageWidget(),
-          Expanded(
-            child: SearchResultWidget(),
-          ),
+          if (searchQuery.query.isEmpty)
+            Expanded(
+              child: SearchInitialPageWidget(),
+            ),
+          if (searchQuery.query.isNotEmpty)
+            Expanded(
+              child: SearchResultWidget(),
+            ),
         ],
       ),
     );
--- a/lib/views/search/search_view.dart
+++ b/lib/views/search/search_view.dart
@@ -1,5 +1,6 @@
 import 'package:flutter/material.dart';
 import 'widgets/search_appbar_widget.dart';
+import 'widgets/search_initial_page_widget.dart';
 import 'widgets/search_result_widget.dart';
 
 class SearchView extends StatelessWidget {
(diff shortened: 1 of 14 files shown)


lib/views/search/search_view.dart
```diff
@@ -15,9 +20,14 @@ class SearchView extends StatelessWidget {
...
       body: Column(
         children: [
-          Expanded(
-            child: SearchResultWidget(),
-          ),
+          if (searchQuery.query.isEmpty)
+            Expanded(
+              child: SearchInitialPageWidget(),
+            ),
+          if (searchQuery.query.isNotEmpty)
```
michaelcspeed: Why not else?

lib/widgets/shimmers/search_initial_page_shimmer_widget.dart
```diff
@@ -0,0 +1,64 @@
...
+import 'package:Medito/constants/constants.dart';
+import 'package:flutter/material.dart';
+
+class SearchInitialPageShimmerWidget extends StatelessWidget {
+  const SearchInitialPageShimmerWidget({super.key});
+
+  @override
+  Widget build(BuildContext context) {
+    return SingleChildScrollView(
+      padding: EdgeInsets.only(top: 20, left: 15, right: 15),
```
michaelcspeed: 15 should probably be 16 here. Can you check please? They will 99% of the time be multiples of 4

lib/widgets/shimmers/search_initial_page_shimmer_widget.dart
```diff
@@ -0,0 +1,64 @@
...
+        mainAxisAlignment: MainAxisAlignment.spaceBetween,
+        crossAxisAlignment: CrossAxisAlignment.start,
+        children: [
+          Expanded(
+            child: Column(
+              crossAxisAlignment: CrossAxisAlignment.start,
+              children: [
+                height8,
+                BoxShimmerWidget(
+                  height: 15,
```
michaelcspeed: These too

lib/views/search/widgets/search_result_widget.dart
```diff
@@ -59,7 +60,13 @@ class SearchResultWidget extends ConsumerWidget {
         onTap: () => ref.refresh(searchProvider),
         isLoading: searchResult.isLoading,
       ),
-      loading: () => const SearchResultShimmerWidget(),
+      loading: () {
+        if (searchQuery.search) {
```
michaelcspeed: Should this be isLoading?

michaelcspeed (changes requested):

lib/views/search/search_view.dart
osamaasifoak: Removed the above condition and instead of this I wrote a ternary condition.

osamaasifoak: `  Expanded(
            child: searchQuery.query.isEmpty
                ? SearchInitialPageWidget()
                : SearchResultWidget(),
          )`

lib/views/search/widgets/search_result_widget.dart
osamaasifoak: This is to check if the user has started a search. The variable name has been changed to hasSearchStarted to avoid confusion.

lib/widgets/shimmers/search_initial_page_shimmer_widget.dart
scouech: Yeah its 16px. If you use the inspect mode on figma you can see the padding is 16px :) thanks both

PR #564, opened 2023-10-28
fix: fade effect added on bg sound

### LINKED ISSUE #557: Make background sound fade out at the end of a track
If background sound is on, we should make it fade out (slowly go from current volume to 0) when the track is about to end. It should only be applied to the background sound.
The previous app had this behaviour. Feel free to ask Mike if you need more info. 
Thank you!

osamaasifoak:
[Make background sound fade out at the end of a track#557](https://github.com/meditohq/medito-app/issues/557)

### DIFF
--- a/lib/providers/player/audio_player_provider.dart
+++ b/lib/providers/player/audio_player_provider.dart
@@ -183,10 +183,21 @@ class AudioPlayerNotifier extends BaseAudioHandler
 
   bool audioPositionIsInEndPeriod(
     Duration position,
-    Duration maxDuration,
-  ) {
-    return maxDuration.inSeconds > 0 &&
+    Duration maxDuration, {
+    bool setToPreviousVolume = false,
+  }) {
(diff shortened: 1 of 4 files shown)


lib/providers/player/audio_player_provider.dart
```diff
@@ -22,7 +22,8 @@ class AudioPlayerNotifier extends BaseAudioHandler
   TrackFilesModel? currentlyPlayingTrack;
   final hasBgSound = 'hasBgSound';
   final trackAudioPlayer = AudioPlayer();
-
+  final fade_duration = 5;
```
michaelcspeed: This should be lowerCamelCase shouldn't it?

michaelcspeed (changes requested):
Just checking, if you start a new session or move the player to a point somewhere in the night of the session after a fade has started, is the volume reset to normal?

osamaasifoak:
Yes, it will reset to its original state.

PR #571, opened 2023-10-30
feat: clear old app downloaded files

### LINKED ISSUE #560: Clear downloaded files / mp3 when first open
When we get the transferStats, we clean the cache, but I don't think that's cleaning the downloaded files (mp3) from the previous version, is it? We should clear the local storage also. Can we do that if it's not already the case? 
Cheers

osamaasifoak:
[Clear downloaded files / mp3 when first open#560](https://github.com/meditohq/medito-app/issues/560)

### DIFF
--- a/lib/repositories/downloader/downloader_repository.dart
+++ b/lib/repositories/downloader/downloader_repository.dart
@@ -1,9 +1,14 @@
+import 'dart:convert';
 import 'dart:io';
 
+import 'package:Medito/constants/strings/shared_preference_constants.dart';
+import 'package:Medito/providers/providers.dart';
 import 'package:Medito/services/network/dio_api_service.dart';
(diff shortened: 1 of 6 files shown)


lib/repositories/downloader/downloader_repository.dart
```diff
@@ -69,6 +75,35 @@ class DownloaderRepositoryImpl extends DownloaderRepository {
...
+            var json = jsonDecode(element);
+            var id = json['id'];
+            var filePath = await getFilePathForOldAppDownloadedFiles(id);
+            var file = File(filePath);
+
+            if (await file.exists()) {
+              await file.delete();
+            }
+          } catch (exception, stackTrace) {
+            print(stackTrace);
```
michaelcspeed: Could post something to sentry?

michaelcspeed (approved):

PR #584, opened 2023-11-02
fix: currently playing session fixed

### LINKED ISSUE #583: Error: Unable to load audio. Please go back and try again
Could we investigate why this error message appears after a meditation for this person? 
In which case it happens, and if we can find a fix (maybe also look at sentry?)

Cheers

[https://discord.com/channels/1137078055358173226/1156182132415270912/1169587162233180190](https://discord.com/channels/1137078055358173226/1156182132415270912/1169587162233180190)

osamaasifoak:
[Error: Unable to load audio. Please go back and try again#583](https://github.com/meditohq/medito-app/issues/583)

### DIFF
--- a/lib/providers/root/root_combine_provider.dart
+++ b/lib/providers/root/root_combine_provider.dart
@@ -25,13 +25,14 @@ final rootCombineProvider = Provider.family<void, BuildContext>((ref, context) {
       .distinct();
   streamEvent.forEach((element) {
     if (element == ProcessingState.completed) {
-      _handleAudioCompletion(ref);
+      _handleAudioCompletion(context, ref);
       _handleUserNotSignedIn(ref, context);
     }
   });
 });
 
 void _handleAudioCompletion(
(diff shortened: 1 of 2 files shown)


lib/providers/root/root_combine_provider.dart
```diff
@@ -25,13 +25,14 @@ final rootCombineProvider = Provider.family<void, BuildContext>((ref, context) {
       .distinct();
   streamEvent.forEach((element) {
     if (element == ProcessingState.completed) {
-      _handleAudioCompletion(ref);
+      _handleAudioCompletion(context, ref);
       _handleUserNotSignedIn(ref, context);
```
michaelcspeed, line 29: It's good practice to keep the params in the same order. 

So like this:
` _handleAudioCompletion(context, ref);
      _handleUserNotSignedIn(context, ref);`

michaelcspeed (approved):

PR #595, opened 2023-11-14
Big player refresh

### LINKED ISSUE #574: Big player refresh
The big player has some small UI changes, we should also add the like button.

See figma here:
[https://www.figma.com/file/E9k4sKjlBDUREveJbUtf4A/Medito-2023?type=design&node-id=2308%3A13852&mode=design&t=jBGq5ZOIlloVCPSl-1](https://www.figma.com/file/E9k4sKjlBDUREveJbUtf4A/Medito-2023?type=design&node-id=2308%3A13852&mode=design&t=jBGq5ZOIlloVCPSl-1
)
- [x] The order of the cover image, title etc have changed
- [x] The forward button is now 10s instead of 30
- [x] The bottom button bar has 4 options:
1. Download: white when not downloaded, purple when downloaded. We should trigger a snackbar when the file is downloaded
2. Background sound: white when not on, purple when on. Opens the background sound screen
3. Playback speed: text is white when X.10, and purple when other values, Should loop like current option
4. Like button: white when not liked, red when liked. 

**Like endpoint:**
POST /events
eventName: liked
Body: trackId, audioFileId

To remove a like:
DELETE /events
eventName: liked
Body: trackId, audioFileId

See on postman
[https://medito.postman.co/workspace/Medito~b22b5a3d-4f45-4545-9e0b-5d3d24353fe2/folder/2129110-f0ca036c-6b18-48c4-ac7a-a660c6d1a0b6?ctx=documentation](https://medito.postman.co/workspace/Medito~b22b5a3d-4f45-4545-9e0b-5d3d24353fe2/folder/2129110-f0ca036c-6b18-48c4-ac7a-a660c6d1a0b6?ctx=documentation
)

The `isLiked`field `true` or `false` is returned in the` /tracks/{id}` payload

- [x] FYI: isLiked is also present for the packs in the payload packs/{id}. It's to accommodate liking a pack in the future. If you can make sure we consider that it would be perfect! Cheers

osamaasifoak:
[Big player refresh#574](https://github.com/meditohq/medito-app/issues/574)


lib/providers/meditation/track_provider.dart
```diff
@@ -33,6 +35,56 @@ Future<TrackModel> tracks(
   return trackRepository.fetchTrack(trackId);
 }
 
+@riverpod
+Future<void> likeDislikeTrack(
+  ref, {
+  required bool isLike,
```
michaelcspeed: Should this be `isLiked`?

lib/providers/meditation/track_provider.dart
```diff
@@ -91,3 +143,12 @@ void addCurrentlyPlayingTrackInPreference(
   }
   print(_track);
 }
+
+//ignore: prefer-match-file-name
+class LikeDisLikeModel {
```
michaelcspeed: LikeDislikeModel* (dislike is one word)

lib/views/player/widgets/bottom_actions/bottom_action_widget.dart
```diff
@@ -22,11 +23,12 @@ class BottomActionWidget extends StatelessWidget {
         crossAxisAlignment: CrossAxisAlignment.center,
         mainAxisAlignment: MainAxisAlignment.spaceAround,
         children: [
-          AudioSpeedWidget(),
-          width8,
-          AudioDownloadWidget(
-            trackModel: trackModel,
-            file: file,
+          SizedBox(
+            width: 60,
```
michaelcspeed: Why 60? I thought they were icons evenly spaced out

lib/views/player/widgets/bottom_actions/widgets/audio_speed_widget.dart
```diff
@@ -29,21 +27,28 @@ class _AudioSpeedComponentState extends ConsumerState<AudioSpeedWidget> {
...
-          ? ColorConstants.onyx
-          : ColorConstants.walterWhite,
-      onTap: () {
-        _provider.setAudioTrackSpeed();
-        ref
-            .read(audioPlayerNotifierProvider)
-            .setTrackAudioSpeed(_provider.audioSpeedModel.speed);
-      },
+    return SizedBox(
+      width: 60,
```
michaelcspeed: Same here... Not sure why 60

lib/views/track/track_view.dart
```diff
@@ -102,9 +102,12 @@ class _TrackViewState extends ConsumerState<TrackView>
     return tracks.when(
       skipLoadingOnRefresh: false,
       data: (data) => _buildScaffoldWithData(context, data, ref),
-      error: (err, stack) => MeditoErrorWidget(
-        message: err.toString(),
-        onTap: () => ref.refresh(tracksProvider(trackId: widget.id)),
+      error: (err, stack) => SizedBox(
+        height: MediaQuery.of(context).size.height / 2,
```
michaelcspeed: Would it be better to wrap the content rather than take up half the screen? Not sure what is happening here so just wondering

michaelcspeed (changes requested):

lib/views/player/widgets/bottom_actions/bottom_action_widget.dart
osamaasifoak: The icons were moving left and right due to variable text width. So, I thought providing a fixed width will solve this issue.

lib/providers/meditation/track_provider.dart
osamaasifoak: Ahh

lib/providers/meditation/track_provider.dart
osamaasifoak: Correct

lib/views/track/track_view.dart
osamaasifoak: The MeditoErrorWidget is a separate widget wrapped with a Scaffold, taking the complete screen height. So, in the bottom sheet, it's opening the bottom sheet on full screen. So, I thought in the error case it would be good if we can show the error in half screen because it's a bottom sheet. Lmk, if you have any suggestion ;)

michaelcspeed: The error widget probably shouldn't have a scaffold then, since it isn't necessarily tied to a screen. I would make it a standalone widget if possible

lib/views/player/widgets/bottom_actions/bottom_action_widget.dart
michaelcspeed: Hardcoding widths like this is a codesmell since it could potentially cause issues on different size phones, or if the user has a custom font size set.
Instead can we use Expanded? I think that will evenly distribute them

osamaasifoak: Hardcoded width removed thanks for the help man

lib/views/track/track_view.dart
osamaasifoak: For now I added a check for scaffold lmk if it's looking good to you.

PR #604, opened 2023-11-17
Replace appbar with medito large appbar

### LINKED ISSUE #596: Replace background sound app bar with MeditoAppBarLarge
Currently the background sound screen is using  `CollapsibleHeaderWidget` could we replace it with the `MeditoAppBarLarge` so that we have only one app bar? 

While doing that, could we make sure that if the cover image is null, not present or returns 4xx or 5xx we don't display it and the app bar is just the background color? 

Cheers

- [x] Replace `collapsibleHeaderWidget` with `MeditoAppBarLarge`
- [x] Handle cover image non present in payload (use color instead)
- [x] Handle cover image returning 4xx (use color instead)
- [x] Handle cover image returning 5xx (use color instead)
- [x] Remove all code related to `CollapsibleHeaderWidget`

osamaasifoak:


lib/views/background_sound/background_sound_view.dart
```diff
@@ -127,30 +134,36 @@ class _BackgroundSoundViewState extends ConsumerState<BackgroundSoundView> {
...
-                .toList(),
+          SliverList(
+            delegate: SliverChildListDelegate([
+              VolumeSliderWidget(),
+              SoundListTileWidget(
+                sound: BackgroundSoundsModel(
+                  id: '0',
+                  title: StringConstants.none,
+                  duration: 0,
+                  path: '',
```
michaelcspeed: Is it possible for these values to be defaults? @osamaasifoak  I was a bit confused seeing the 0s and empty strings

michaelcspeed (approved):

lib/views/background_sound/background_sound_view.dart
osamaasifoak: Fixed!!

lib/views/background_sound/background_sound_view.dart
```diff
@@ -168,4 +161,18 @@ class _BackgroundSoundViewState extends ConsumerState<BackgroundSoundView> {
...
+
+    return SoundListTileWidget(
+      sound: BackgroundSoundsModel(
+        id: noneId,
+        title: StringConstants.none,
+        duration: noneDuration,
+        path: '',
+      ),
+    );
+  }
```
michaelcspeed: Sorry I meant defaults of BackgroundSoundsModel.  Do you think that's a good idea?

michaelcspeed: This is good though too, much clearer that is the one for None :)

michaelcspeed (approved):

PR #610, opened 2023-11-22
UI player

scouech:


scouech:
@michaelcspeed if you could review this one too please 👼 Could you check if everything is alright with the back gesture? I've added it, it should close the audio player. 
Also, if you can make sure everything is correct with the playback speed please!
Thank you

michaelcspeed (approved):

PR #621, opened 2023-11-28
Tidy up code

michaelcspeed:
Made some improvements to readability.  @osamaasifoak lemme know if they make sense to you

### DIFF
--- a/lib/providers/player/audio_player_provider.dart
+++ b/lib/providers/player/audio_player_provider.dart
@@ -23,7 +23,7 @@ class AudioPlayerNotifier extends BaseAudioHandler
   final trackAudioPlayer = AudioPlayer();
 
   TrackFilesModel? currentlyPlayingTrack;
-  final hasBgSoundId = 'hasBgSound';
+  final hasBgSoundKey = 'hasBgSound';
   final fadeDurationInSeconds = 5;
   var bgVolume;
 
@@ -210,7 +210,7 @@ class AudioPlayerNotifier extends BaseAudioHandler
         trackModel.coverUrl,
       ),
       extras: {
-        hasBgSoundId: trackModel.hasBackgroundSound,
+        hasBgSoundKey: trackModel.hasBackgroundSound,
         'trackId': trackModel.id,
         'fileId': file.id,
       },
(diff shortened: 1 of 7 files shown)


lib/providers/player/player_provider.dart
```diff
@@ -14,100 +14,99 @@ final playerProvider =
...
     required TrackFilesModel file,
   }) async {
-    var _track = trackModel.customCopyWith();
-    for (var i = 0; i < _track.audio.length; i++) {
-      var element = _track.audio[i];
-      var fileIndex = element.files.indexWhere((e) => e.id == file.id);
+    var track = trackModel.customCopyWith();
+
+    track.audio.forEach((audioModel) {
+      var fileIndex = audioModel.files.indexWhere((it) => it.id == file.id);
```
osamaasifoak, line 25: what's the meaning of **it**?

lib/providers/player/audio_player_provider.dart
```diff
@@ -19,10 +20,11 @@ final audioPlayerNotifierProvider =
 class AudioPlayerNotifier extends BaseAudioHandler
     with QueueHandler, SeekHandler, ChangeNotifier {
   var backgroundSoundAudioPlayer = AudioPlayer();
-  TrackFilesModel? currentlyPlayingTrack;
-  final hasBgSound = 'hasBgSound';
   final trackAudioPlayer = AudioPlayer();
-  final fadeDuration = 5;
+
+  TrackFilesModel? currentlyPlayingTrack;
+  final hasBgSoundId = 'hasBgSound';
```
osamaasifoak: I think the name should be hasBgSound instead of hasBgSoundId. Lmk, what do you think?

osamaasifoak (approved):
Hey @michaelcspeed, I'm approving this PR. The changes you have made are looking good to me thanks man :)

lib/providers/player/audio_player_provider.dart
michaelcspeed: `hasBGSound` implies it's a boolean, but this is really a key... maybe it should be called `hasBgSoundKey`

lib/providers/player/player_provider.dart
michaelcspeed: It's the same as a for loop, but directly gives you the item instead if incrementing an index

lib/providers/player/audio_player_provider.dart
michaelcspeed: Also 'trackId' and 'fileId' should be constants too but I didn't do them here

michaelcspeed:
@osamaasifoak If you're happy with this, please merge it once you've resolved the comments. Thanks!

lib/providers/player/audio_player_provider.dart
osamaasifoak: Correct **hasBgSound** should be called **hasBgSoundKey**

PR #624, opened 2023-12-01
chore: custom headers added, code refactoring

### LINKED ISSUE #622: Add custom header information to all requests
Right now we are doing an "appOpened" event with the following information

`{
    "name": "appOpened",
    "payload": {
        "deviceOs": "iOS",
        "deviceLanguage": "en-US",
        "deviceModel": "iPhone 12 Pro Max",
        "appVersion": "2.1.3"
    }
}`

To make easier to debug on the backend, we need to pass that data in the header of every requests to the backend (POST, GET, DELETE, or whatever other call we are making). 

The information should be in the header formatted like this:

    'Device-Os': '{iOS or Android}',
    'Device-Language': '{in the format like en-US}',
    'Device-Model': '{the model such as iPhone 12 Pro Max}',
    'App-Version': '{the app version}'

We should also add the device time as
` 'Device-Time': '{the device time }'
`
More info here:
https://api.flutter.dev/flutter/dart-core/DateTime-class.html

osamaasifoak:
[Add custom header information to all requests#622](https://github.com/meditohq/medito-app/issues/622)


osamaasifoak:
@michaelcspeed, let me know if you have any suggestions regarding this app initialization provider thanks.

lib/providers/auth/app_initialization_provider.dart
michaelcspeed, line 1: This is very hard to read and understand because it's doing too much.
It's handling auth, connectivity, error handling, and navigation. That's a hefty load for one provider, which will make it hard to maintain and test.
I would remove the snackbar stuff and put it somewhere else. Only show snackbars from UI classes.

I don't think this is the right place to assign the headers, is it? It should be done in dioClientProvider.

lib/providers/auth/app_initialization_provider.dart
```diff
@@ -17,78 +17,29 @@ final appInitializationProvider =
   try {
     var retryCount = ref.read(retryCounterProvider);
     if (retryCount < 2) {
-      _checkInternetConnectivity(context, ref);
+      var connectivityStatus = ref.read(connectivityStatusProvider);
+      if (connectivityStatus == ConnectivityStatus.isDisonnected) {
```
michaelcspeed: There is a typo here. You should turn on the spellchecker in VSCode

lib/providers/auth/app_initialization_provider.dart
```diff
@@ -17,78 +17,29 @@ final appInitializationProvider =
   try {
     var retryCount = ref.read(retryCounterProvider);
     if (retryCount < 2) {
```
michaelcspeed: This 2 should be a constant, and I don't think retryCounterProvider is necessary is it? It can just be a var

lib/providers/auth/assign_dio_headers_provider.dart
```diff
@@ -0,0 +1,30 @@
+import 'dart:io';
+
+import 'package:Medito/models/models.dart';
+import 'package:Medito/providers/providers.dart';
+import 'package:Medito/services/network/dio_client_provider.dart';
+import 'package:flutter_riverpod/flutter_riverpod.dart';
+
+var assignDioHeadersProvider = FutureProvider<void>((ref) async {
+  var _authProvider = ref.read(authProvider);
```
michaelcspeed: If the var is declared inside a function it doesn't need to be private, since the scope is already restricted.

lib/providers/auth/auth_retry_provider.dart
```diff
@@ -0,0 +1,45 @@
...
+import 'package:flutter_riverpod/flutter_riverpod.dart';
+import 'package:go_router/go_router.dart';
+
+import 'app_initialization_provider.dart';
+
+final authRetryProvider =
+    StateProvider.family<void, AuthRetryModel>((ref, data) {
+  var retryCount = ref.watch(retryCounterProvider);
+
+  if (retryCount < 2) {
```
michaelcspeed: Calling retryCount in 2 places should be a red flag :D

lib/providers/auth/auth_retry_provider.dart
```diff
@@ -0,0 +1,45 @@
...
+    Duration(seconds: 2),
+    () => ref.refresh(appInitializationProvider(context)),
+  );
+}
+
+//ignore: prefer-match-file-name
+class AuthRetryModel {
+  BuildContext context;
+  bool isNavigate;
+  AuthRetryModel({required this.context, required this.isNavigate});
```
michaelcspeed: It's a good idea to separate UI and business logic like I said above. We don't want to tightly couple the two because it makes things more difficult in the long run. So maybe consider observing a state from the UI, and navigating when it changes. That means you probably don't even need this auth_retry_provider

lib/providers/auth/app_initialization_provider.dart
```diff
@@ -17,78 +17,29 @@ final appInitializationProvider =
...
-      _checkInternetConnectivity(context, ref);
+      var connectivityStatus = ref.read(connectivityStatusProvider);
+      if (connectivityStatus == ConnectivityStatus.isDisonnected) {
+        showSnackBar(context, StringConstants.connectivityError);
+      }
     }
     await _authProvider.initializeUser();
     var res = _authProvider.userRes;
```
michaelcspeed: Can we try to avoid abbreviations? Res could be resource, response etc. It's a good idea to keep in mind that the code needs to be immediately understandable to anyone who's never seen it before :)

lib/providers/auth/initialize_user_provider.dart
```diff
@@ -0,0 +1,17 @@
+import 'package:Medito/providers/providers.dart';
+import 'package:Medito/services/network/dio_client_provider.dart';
+import 'package:flutter_riverpod/flutter_riverpod.dart';
+
+final initializeUserProvider = FutureProvider<bool>((ref) async {
+  var auth = ref.read(authProvider);
+  await auth.initializeUser();
+  var response = auth.userRes;
```
michaelcspeed: `userRes` is an abbreviation

lib/providers/events/app_opened_event_provider.dart
```diff
@@ -0,0 +1,27 @@
...
+import 'package:flutter_riverpod/flutter_riverpod.dart';
+
+var appOpenedEventProvider = FutureProvider<void>((ref) async {
+  var deviceInfo = await ref.read(deviceAndAppInfoProvider.future);
+  var data = _setAppOpenedModelData(deviceInfo);
+
+  await ref.read(eventsProvider(event: data.toJson()).future);
+});
+
+EventsModel _setAppOpenedModelData(DeviceAndAppInfoModel val) {
```
michaelcspeed: Is this a setter or a getter?

lib/services/network/dio_client_provider.dart
```diff
@@ -62,3 +64,27 @@ Future<void> _captureException(
     stackTrace: err.stackTrace,
   );
 }
+
+var assignDioHeadersProvider = FutureProvider<void>((ref) async {
+  var _authProvider = ref.read(authProvider);
```
michaelcspeed: `_`!

lib/services/network/dio_client_provider.dart
```diff
@@ -62,3 +64,27 @@ Future<void> _captureException(
...
+  var deviceInfo = await ref.read(deviceAndAppInfoProvider.future);
+  var headers = ref.read(dioClientProvider).dio.options.headers;
+
+  var _user = _authProvider.userRes.body as UserTokenModel;
```
michaelcspeed: Here too...

lib/views/background_sound/background_sound_view.dart
```diff
@@ -61,12 +61,10 @@ class _BackgroundSoundViewState extends ConsumerState<BackgroundSoundView> {
...
   @override
   Widget build(BuildContext context) {
-    var connectivityStatus =
-        ref.watch(connectivityStatusProvider) as ConnectivityStatus;
+    var connectivityStatus = ref.watch(connectivityStatusProvider);
 
     ref.listen(connectivityStatusProvider, (prev, next) {
-      var state = next as ConnectivityStatus;
-      if (state == ConnectivityStatus.isDisonnected) {
+      if (next == ConnectivityStatus.isDisonnected) {
```
michaelcspeed: This is still spelled wrong.

lib/views/splash_view.dart
```diff
@@ -12,16 +14,53 @@ class SplashView extends ConsumerStatefulWidget {
...
+      } else {
+        retryInitializeUser();
+      }
+    } catch (e) {
+      retryInitializeUser();
+    }
+  }
+
+  void retryInitializeUser() {
+    if (retryCounter < maxRetryCount) {
```
michaelcspeed: This retry counter is business logic and shouldn't be in the view. The Splash screen doesn't need to know anything about the retry logic.

michaelcspeed:
This looks a lot better though, good job. Just a few little things to fix

lib/providers/events/app_opened_event_provider.dart
osamaasifoak: This is only for creating a model I changed the name to _createAppOpenedModelData

lib/views/background_sound/background_sound_view.dart
osamaasifoak: Ahh fixed

michaelcspeed:
Thanks for the improvements @osamaasifoak !

PR #626, opened 2023-12-08
Home screen UI

### LINKED ISSUE #623: Home revamp
Based on the (very) high number of feedback we have received, we are revamping the home screen. This will not only help regarding the feedback, but also allow us to allocate the quote, stats in a more prominent way and other media items we want to display. 

# Design
You can find the design in this new figma file:
https://www.figma.com/file/EjZtnDvYjrWKaYKRQpU2oq/Medito-2023-v2?type=design&node-id=2805%3A15569&mode=design&t=H0UrW3O32HTLur9v-1

# Endpoint
The home screen is now divided in multiple parts, each part has its own endpoint:
## Header
`GET /main/header`
Basically the top bar navigation. It has three elements:
1. Greeting
2. Announcement (same as we currently have)
3. Menu (same as we currently have)

- The announcement can be present, if dismissed it will not be in the payload
- Greeting can change based on the payload

## Shortcuts
`GET /main/shortcuts`
It's a section with shortcuts, similar to the chips we currently have
- People should be able to reorder them by drag/drop. It should be saved locally for now

## Quote
`GET /main/quote`
It shows a quote with text and an author. The author is below the text.

## Tiles: stats
`GET /main/stats`
For now we have two tiles:
1. A first tile, can be any sort of data but will always be the same format
a. An icon
b. A title
c. A subtitle
d. A color, for the icon color
The first tile should open the bottom sheet with the stats. The stats are in the payload under "allStats"

2. A second tile, can be any sort of data, always the same format (same as above)
There is nothing to open here.

osamaasifoak:
[Home revamp#623](https://github.com/meditohq/medito-app/issues/623)


lib/views/home/home_view.dart
```diff
@@ -21,151 +21,95 @@ class HomeView extends ConsumerStatefulWidget {
...
+  void onPointerDown(_) async {
+    if (await Vibration.hasVibrator() ?? false) {
+      if ((await Vibration.hasAmplitudeControl()) ?? false) {
+        unawaited(Vibration.vibrate(amplitude: 20, duration: 30));
+      } else {
+        unawaited(Vibration.vibrate(duration: 30));
+      }
+    }
+  }
+
```
michaelcspeed: Let's remove this vibration thing. I spoke to @scouech about it. It's really different on all devices, some devices have a really weak vibration and some a really strong one so it's not a great UX.

lib/views/home/widgets/header_and_announcement_widget.dart
```diff
@@ -0,0 +1,99 @@
...
+
+  @override
+  ConsumerState<HeaderAndAnnouncementWidget> createState() =>
+      _HeaderAndAnnouncementWidgetState();
+}
+
+class _HeaderAndAnnouncementWidgetState
+    extends ConsumerState<HeaderAndAnnouncementWidget>
+    with TickerProviderStateMixin, AutomaticKeepAliveClientMixin {
+  bool _isCollapsed = false;
```
michaelcspeed: don't need the underscore here (see your last PR)

lib/views/home/widgets/shortcuts/shortcuts_widget.dart
```diff
@@ -0,0 +1,126 @@
...
+class ShortcutsWidget extends ConsumerStatefulWidget {
+  const ShortcutsWidget({super.key});
+
+  @override
+  ConsumerState<ShortcutsWidget> createState() => _ShortcutsWidgetState();
+}
+
+class _ShortcutsWidgetState extends ConsumerState<ShortcutsWidget> {
+  late ShortcutsModel data;
+  @override
```
michaelcspeed: Can we try to space things out a bit better? Always have a line before a method for example. Sorry to be pedantic but it really helps me with readability. Thanks!

lib/views/search/widgets/search_initial_page_widget.dart
```diff
@@ -40,8 +40,8 @@ class SearchInitialPageWidget extends ConsumerWidget {
               keyboardDismissBehavior: ScrollViewKeyboardDismissBehavior.onDrag,
               padding: EdgeInsets.only(
                 top: 20,
```
michaelcspeed: Maybe we should have a padding20

lib/widgets/buttons/loading_button_widget.dart
```diff
@@ -32,7 +32,7 @@ class LoadingButtonWidget extends StatelessWidget {
       disabledColor: bgColor.withOpacity(0.7),
       color: bgColor,
       splashColor: ColorConstants.transparent,
-      padding: EdgeInsets.symmetric(vertical: 12, horizontal: defaultPadding),
+      padding: EdgeInsets.symmetric(vertical: 12, horizontal: padding16),
```
michaelcspeed: and a padding 12?

lib/models/home/shortcuts/shortcuts_model.dart
```diff
@@ -0,0 +1,27 @@
+import 'package:freezed_annotation/freezed_annotation.dart';
+
+part 'shortcuts_model.freezed.dart';
+part 'shortcuts_model.g.dart';
+
+@unfreezed
```
michaelcspeed, line 6: Just wondering why these need to be mutable?

lib/views/home/home_view.dart
osamaasifoak: Removed

lib/views/search/widgets/search_initial_page_widget.dart
osamaasifoak: Correct!!

lib/models/home/shortcuts/shortcuts_model.dart
osamaasifoak: It is because while reordering the item in grid view, it threw me an error.

michaelcspeed: Ah, ok that makes sense, but I think it exposes a weakness in our architecture. We're currently using the same model for both backend data retrieval and UI state management. This approach could lead to some challenges:

**Immutable vs Mutable:** Our backend model, ideally, should be immutable to maintain a single source of truth. Making it mutable for UI purposes can introduce unpredictable state changes and complexity.

**Separation of Concerns:** By using the same model for both backend and UI, we're blurring the lines between our data layer and presentation layer. This goes against the separation of concerns principle, potentially making our codebase harder to maintain.

**State Management:** With our current setup, any changes in the UI directly affect our model state. In a reactive framework like Flutter, coupled with Riverpod, this might lead to unintended side effects and hard-to-track bugs.

I’ve also noticed that we have a similar implementation elsewhere in the codebase (TrackModel?). My suggestion would be to differentiate between the backend model and the UI model. We could keep our backend model (@freezed and immutable) to represent the data as it is in the backend. For the UI, we can introduce a separate, mutable model that reflects the user interactions and UI state. This approach would not only improve the maintainability and predictability of our app but also aligns better with best practices in app architecture.

I'll create a ticket for us to review and update the other instances where we've used @unfreezed, and we can do it in another PR.

wdyt?

PR #630, opened 2023-12-11
fix: track duration dropdown fixed

### LINKED ISSUE #625: Not all durations are showing per track
If you call 
`/tracks/5dZCRQnJ75PN1aB4`
(Which is the track called "Anxiety")
You can see 3 guides in the response, with 4 mp3 in total. 1 for Will, 1 for Steven and 2 for Regina. However the app only shows one option per guide. The one from Regina with id `FP5Kmz2LWvh87n0L` is missing. 

It seems that the issue also happens for other tracks, like `zHNJ1zL04PMAxzRF`. I suppose the UI doesn't display all the options per guide (note that each guide can have different durations)

osamaasifoak:
[Not all durations are showing per track#625](https://github.com/meditohq/medito-app/issues/625)


lib/views/track/track_view.dart
```diff
@@ -90,11 +90,15 @@ class _TrackViewState extends ConsumerState<TrackView>
...
-    unawaited(context.push(RouteConstants.playerPath));
+    try {
+      await ref.read(playerProvider.notifier).loadSelectedTrack(
+            trackModel: trackModel,
+            file: file,
+          );
+      unawaited(context.push(RouteConstants.playerPath));
+    } catch (e) {
+      print(e);
+    }
```
michaelcspeed: Why are we just swallowing this exception? What kind of exception might we expect?

lib/views/track/track_view.dart
```diff
@@ -268,16 +267,16 @@ class _TrackViewState extends ConsumerState<TrackView>
       topRight: 7,
       bottomRight: 7,
       bottomLeft: 7,
-      isDisabled: audioFiles.length > 1,
+      isDisabled: (_selectedFile ?? audioFiles).length > 1,
```
michaelcspeed: can remove the _

lib/views/track/track_view.dart
osamaasifoak: I apologize for adding this block; it was only for testing purposes.

PR #632, opened 2023-12-11
Remove try/catch with just rethrow

michaelcspeed:
In this PR, I've stripped out several try/catch blocks that were only using rethrow and nothing else. These were kind of pointless – they caught exceptions only to throw them again without any additional handling or logging. 

Removing them makes the code leaner and more readable, without affecting its functionality. If we're not doing anything special with the caught exceptions, there's no need to catch them in the first place. :)


michaelcspeed:
I'm not sure why it thinks there are so many changes... they should have been merged in the other PR.

But the relevant commit is here https://github.com/meditohq/medito-app/pull/632/commits/2b6abdfd994c02454e03555ed5af7cc1a2547818

PR #639, opened 2023-12-14
Fix home screen

### LINKED ISSUE #637: Fix home screen
1. When you pull down to refresh it flickers

> - could we make sure the shimmer view is more or less correct, so it should basically fix the issue?

> - If not, then please tell me, but we will need a better way to display the refreshed content

2. What’s the weird scrollview height thingy? If you pull down to refresh (like keep pulling down) there is some kind of container, like the view should fill up the entire screen, it’s not the case atm (see screenshot)

3. When dragging, there is some kind of background on the shorcut, could we make sure the dragged item has either no dark background, or a border radius so it’s not visible?

Image

Image

osamaasifoak:
[Fix home screen#637](https://github.com/meditohq/medito-app/issues/637)


lib/views/home/home_view.dart
```diff
@@ -41,7 +41,7 @@ class _HomeViewState extends ConsumerState<HomeView>
             children: [
               Expanded(
                 child: RefreshIndicator(
-                  onRefresh: _onRefresh,
+                  onRefresh: onRefresh,
```
michaelcspeed: This should be private shouldn't it?

lib/views/home/home_view.dart
```diff
@@ -73,20 +73,18 @@ class _HomeViewState extends ConsumerState<HomeView>
...
     ref.invalidate(fetchQuoteProvider);
     await ref.read(fetchQuoteProvider.future);
-    ref.invalidate(fetchHomeHeaderProvider);
-    await ref.read(fetchHomeHeaderProvider.future);
+    ref.invalidate(remoteStatsProvider);
+    await ref.read(remoteStatsProvider.future);
   }
 
-  FloatingActionButton _buildFloatingButton(BuildContext context) {
+  FloatingActionButton buildFloatingButton(BuildContext context) {
```
michaelcspeed: This too

lib/views/home/widgets/shortcuts/shortcuts_widget.dart
```diff
@@ -17,25 +17,72 @@ class ShortcutsWidget extends ConsumerStatefulWidget {
...
 class _ShortcutsWidgetState extends ConsumerState<ShortcutsWidget> {
   late ShortcutsModel data;
-
+  late List<Widget> shortcutsWidgetList = [];
   @override
   void initState() {
     super.initState();
   }
 
+  void handleChipPress(
```
michaelcspeed: All these methods here should be private

lib/views/home/widgets/tiles/tiles_widget.dart
```diff
@@ -30,8 +30,7 @@ class TilesWidget extends ConsumerWidget {
     );
   }
 
-  Padding _buildTiles(
-    BuildContext context,
+  Padding buildTiles(
```
michaelcspeed: This too... Why are they all public now?

lib/views/home/widgets/tiles/tiles_widget.dart
```diff
@@ -92,6 +99,16 @@ class TilesWidget extends ConsumerWidget {
     );
   }
 
+  double getFontSize(BoxConstraints constraints) {
+    final width = constraints.maxWidth;
+    var fontSize = 16.0;
+    if (width <= 480) {
```
michaelcspeed: Magic number?

michaelcspeed (changes requested):

lib/views/home/home_view.dart
osamaasifoak: Yeah this should be private damn it remove underscore from all the places lol let me fix it real quick

lib/views/home/widgets/tiles/tiles_widget.dart
osamaasifoak: Sorry that was my fault

michaelcspeed (approved):

PR #654, opened 2023-12-26
Show download on 5xx

### LINKED ISSUE #636: If error from the server (5xx), then show "Downloads" screen
At the moment the app displays this message if the server is not reachable (error 5xx). We should direct people to the Downloads screen instead. 
Is that possible?

Maybe we could display the "retry" button and the "Downloads" button, and improve this error message? 

Image

osamaasifoak:
[If error from the server (5xx), then show "Downloads" screen#636](https://github.com/meditohq/medito-app/issues/636)


lib/constants/strings/string_constants.dart
```diff
@@ -23,10 +23,11 @@ class StringConstants {
   static const String wellDoneCopy = 'Well done for meditating today!';
   static const String wellDoneSubtitle =
       'The mind is everything. What you think, you become.';
-  static const String meanWhileCheck = 'Meanwhile, you can check';
+  static const String meanWhileCheck = 'Meanwhile, you can listen to your';
```
michaelcspeed: The name of this const is inaccurate now

lib/widgets/errors/medito_error_widget.dart
```diff
@@ -21,26 +21,42 @@ class MeditoErrorWidget extends StatelessWidget {
 
   @override
   Widget build(BuildContext context) {
-    var isInvalidToken = message == StringConstants.invalidToken;
+    var splittedMessage = message.split(': ');
+    var _showCheckDownloadText = showCheckDownloadText;
 
+    if (splittedMessage.length > 1) {
+      var statusCode = int.parse(splittedMessage[1]);
```
michaelcspeed: This could throw an exception. I would split all this logic out into another method and also handle the exception.

lib/widgets/errors/medito_error_widget.dart
```diff
@@ -21,26 +21,42 @@ class MeditoErrorWidget extends StatelessWidget {
 
   @override
   Widget build(BuildContext context) {
-    var isInvalidToken = message == StringConstants.invalidToken;
+    var splittedMessage = message.split(': ');
+    var _showCheckDownloadText = showCheckDownloadText;
```
michaelcspeed: This could cause some confusion. If someone sets `MeditoErrorWidget(showCheckDownloadText:false)` but then the message contains a 500 error, the Download button will appear. It might be best to
- get rid of the showCheckDownloadText
- clarify that showCheckDownloadText doesn't really guarantee that it will show (not great in my opinion! :D)
- **or** take the statusCode stuff out of this class and check it somewhere else. This is probably the best option because the view should not need to know anything about the network layer.

michaelcspeed: I asked ChatGPT about it: 

Yes, it's generally a good practice to separate concerns in your application, especially between the view (UI) and the business logic. In the context of your MeditoErrorWidget, handling status codes and other logic within the widget can lead to a couple of issues:

**Coupling:** Your UI components become tightly coupled with business logic, making it harder to reuse, test, and maintain them. Ideally, widgets should be focused on presenting data and handling user interactions, not on processing or interpreting the data.

**Readability and Maintainability:** Mixing business logic with UI code can make your codebase more complex and harder to read. Keeping these concerns separate helps maintain a clean and understandable code structure.

**Scalability:** As your application grows, having business logic in your UI layers can make it difficult to manage and scale the application. Changes in business logic would require changes in the UI layer, which is not ideal.

**Testing:** It's easier to write unit tests for business logic when it's decoupled from the UI. If the logic is embedded in the UI, testing becomes more challenging and less reliable.

To improve the design, consider the following approach:

**ViewModel or Controller:** Implement a ViewModel or Controller that handles the business logic, such as interpreting status codes. This component can process the data and then provide the necessary information to the UI in a simple and straightforward manner.

**Data Passing:** Pass only the necessary data to the widget. For example, instead of passing the entire error message with the status code, you could pass a simplified error message or an enum indicating the type of error.

**UI Focused Widgets:** Ensure that your widgets remain focused on presenting data and handling user interactions. Any logic related to data processing or interpretation should be handled outside the widget.

Refactoring your widget to follow these guidelines will not only make your code cleaner and more maintainable, but also prepare it for future changes and enhancements.

osamaasifoak: Hi @michaelcspeed, I can separate the message splitting logic and I will also handle the exception. The problem is with the **showCheckDownloadText**. According to our requirement if the status code is 5xx we need to show the check download text but we also need to show the check download text in the case of the connectivity error. I tested this thing I generated the 5xx error and then connectivity error and it's working fine.  And I think it's not directly communicating with the network layer because here we only checking the status code in the message returned from the network layer let me show you the hierarchy.
API services => Repository => Provider => UI

michaelcspeed: hey @osamaasifoak Thanks for addressing the issue and explaining the current implementation. :) I understand the need to show `checkDownloadText` for both 5xx errors and connectivity errors. My concern is primarily about keeping our UI components as simple and focused on presentation as possible.

Would it be feasible to move the logic for determining when to show checkDownloadText into the Provider layer? For example, these layers could process the error and set a flag or an error type that the UI simply reads and acts upon. This way, our UI components won't need to interpret the error message or status codes directly.

This approach could help us make our UI components more reusable and easier to maintain, as they wouldn't be tied to specific logic about error handling. It will also make unit testing of UI components and business logic separately more straightforward.

Happy to discuss further if it needs more clarification. Thanks!

osamaasifoak: Hi @michaelcspeed, I updated the code as per the above instruction let me know if it's okay now thanks

michaelcspeed:
Cool, this is better but we still have the problem where `isShowCheckDownload` could be set to false but depending on the message it could still show. I would make it so that the logic is done _outside_ the view, and you pass through the UIState class as a field.

lib/providers/error_widget/medito_error_widget_provider.dart
```diff
@@ -0,0 +1,31 @@
...
+
+final meditoErrorWidgetProvider =
+    Provider.family<MeditoErrorWidgetUIState, MeditoErrorWidgetUIState>(
+  (ref, data) {
+    try {
+      if (data.shouldShowCheckDownload) {
+        return MeditoErrorWidgetUIState(true, data.message);
+      }
+      var splittedMessage = data.message.split(': ');
+      if (splittedMessage.length > 1) {
```
michaelcspeed: 'splitted' isn't standard English - should be 'split'

lib/providers/error_widget/medito_error_widget_provider.dart
```diff
@@ -0,0 +1,31 @@
...
+      return MeditoErrorWidgetUIState(false, splittedMessage[0]);
+    } catch (e) {
+      return MeditoErrorWidgetUIState(false, data.message);
+    }
+  },
+);
+
+//ignore: prefer-match-file-name
+class MeditoErrorWidgetUIState {
+  final bool shouldShowCheckDownload;
```
michaelcspeed: might be clearer to call it `shouldShowCheckDownloadButton`

michaelcspeed:
> Cool, this is better but we still have the problem where `isShowCheckDownload` could be set to false but depending on the message it could still show. I would make it so that the logic is done _outside_ the view, and you pass through the UIState class as a field.

@osamaasifoak Not sure if you missed this comment

PR #655, opened 2023-12-26
feat: support added for non google play service devices

### LINKED ISSUE #581: [BUG] Version 3.0 does not work without Google Play Services on Android
Version 3.0 does not work without Google Play Services on Android. Instead, the app displays a blank screen. This issue makes the app unusable on Android-based devices that do not include Google Play Services, including Amazon's Fire tablets, devices sold in places where Google is banned (China), and privacy-oriented devices that exclude Google services.

This issue occurs because the app refuses to load any content if Firebase Cloud Messaging (implemented in #373) fails to generate a registration token for the device. To fix the issue and enable all Android devices to access Medito, the app should continue to show content even if FCM is unavailable. (For devices without Google Play Services, the app should not prompt the user to allow push notifications or to create an account, since these features are currently implemented in a way that depends on FCM.)

osamaasifoak:
[[BUG] Version 3.0 does not work without Google Play Services on Android#581](https://github.com/meditohq/medito-app/issues/581)

### DIFF
--- a/lib/utils/utils.dart
+++ b/lib/utils/utils.dart
@@ -13,6 +13,7 @@ Affero GNU General Public License for more details.
 You should have received a copy of the Affero GNU General Public License
 along with Medito App. If not, see <https://www.gnu.org/licenses/>.*/
 
+import 'dart:async';
 import 'dart:io';
 import 'dart:typed_data';
 import 'dart:ui' as ui;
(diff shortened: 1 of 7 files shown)


lib/utils/utils.dart
```diff
@@ -194,6 +195,20 @@ int formatIcon(String icon) {
   return int.parse('0x$icon');
 }
 
+Future<bool> checkGooglePlayServices() async {
+  if (Platform.isIOS) {
+    return true;
+  } else if (Platform.isAndroid) {
+    var availability = await GoogleApiAvailability.instance
+        .checkGooglePlayServicesAvailability();
```
michaelcspeed: Is it possible for this to throw an exception? In the [example](https://pub.dev/packages/google_api_availability/versions) it looks like they are catching one.

osamaasifoak: It appears that they are only catching the platform exception, which is only occurring for iOS devices. I have already added a check to deal with it, but I still wrapped my code inside a try/catch block.

PR #677, opened 2024-02-08
Vectorize android icons

Lionelf329:
Now that the min sdk has been raised to 26, it's no longer necessary to use rasterized icons. I've manually converted ic_launcher_foreground, ic_launcher_background, and ic_notification_icon to vector graphics, and deleted the old png files.

I was careful to convert them as accurately as possible. The launcher background and notification icon are both perfect, and the launcher foreground is close enough that you'll only see a difference if you overlay it with the original png and scale it up to fill your whole monitor. I was also careful to avoid the common pitfall in svg paths where misaligned control points create points of inflection, so there are no sharp angles anywhere. It's also a very efficient path compared to what you'd normally get when trying to turn a png into a vector, so it'll be very performant to render.


michaelcspeed:
This is great, thanks @Lionelf329 !

PR #694, opened 2024-09-12
Enabling iPad version

luz-ams:
I've enabled the iPad configuration in Xcode so we can target this devices as well. 

From this:

IMG_0573

To this:

IMG_0574
IMG_0575
IMG_0576
IMG_0577
IMG_0578
IMG_0579
IMG_0580


michaelcspeed:
Amazing @luz-ams thanks!

PR #695, opened 2024-09-18
Delete dead code and assets

Lionelf329:
There's a lot of unreachable stuff scattered throughout the repo, and it makes it hard to figure out which parts are actually important. This PR deletes a lot of unused code and assets to make it easier to understand what's going on.

I made sure that the app still compiles after my changes, but I didn't do any other testing because I'm assuming that any mistakes I might have made will be caught during compilation. The only thing that might be a problem is if there's some way to use an `assets` image without the literal text "assets" appearing anywhere nearby, but I'm pretty sure that doesn't happen anywhere.


michaelcspeed:
Thanks for this @Lionelf329 I will take a look today. I didn't see it until now, I guess I didn't get a notification because I wasn't tagged :)

michaelcspeed (approved):

michaelcspeed:
Thanks @Lionelf329 !

PR #698, opened 2024-09-24
Improve formatting, remove unused multidex dependency

Lionelf329:
Multidex is only needed for android apps with a min sdk below 20, and ours is 26, so it's safe to remove.

The formatting changes were part of what I was working on yesterday, but I kept it separate from the massive PR because these aren't just whitespace changes and I didn't want to accidentally sneak bugs in by including code changes in what was otherwise mostly a formatting PR. The code was perfectly readable before, but `dart format` was going to butcher it so I moved things around to make it look nice while following the formatting rules.


lib/services/network/assign_dio_headers.dart
michaelcspeed, line 1: I prefer the previous implementation. Here it looks like you are creating a copy of the headers and only setting it locally (I know that's not how it works but that's how it reads unfortunately). The print can be removed though :)

Lionelf329: Okay, no problem! I've updated it to delete the print statement, and added the new autoformat changes from today

michaelcspeed (approved):
Great thank you @Lionelf329

PR #702, opened 2024-10-20
Update README.md ## License

lambdakilo:
Before accepting/merging we should maybe take a look under which CC license the content was released upon.


lambdakilo:
According to Wayback Machine the content was released under CC-BY-NC-SA-4.0 [1]. Where can we see as users which content is still usable under the aforementioned CC license and which content is only released under the custom license?

1: http://web.archive.org/web/20240327141940/https://meditofoundation.org/license

scouech:
Thanks for your input. While we initially considered releasing the content under the CC-BY-NC-SA-4.0 license, we ultimately did not proceed with that plan. The content has never been officially released under that license.

lambdakilo:
I see. Then I'm personally good with the changes in this PR as is.

PR #705, opened 2025-12-01
Upgrade haptic feedback

nohli:
I saw that this is a popular app using my package `haptic_feedback`.

Recently I've made some substantial improvements which will come with this PR:

- Android 13+ users who have disabled touch haptics in system settings didn't get haptics from the app. This change sets the usage type to `media` (before it was `unknown`, which defaults to `touch`). If vibrations are indeed supposed to be disabled when the user opts out from touch haptics, we can remove the usage argument, or set it to touch.

- On Android 11+ the plugin can now use HD vibrations which feel much better and smoother

- Android 7.* devices now vibrate with the correct pattern

https://pub.dev/packages/haptic_feedback/changelog


michaelcspeed:
Thanks @nohli !

nohli:
@michaelcspeed it looks like that in your last [merge commit](https://github.com/meditohq/medito-app/pull/705/changes/b105e22488d4a6940d5f69ac9c3dfde48cd3f7f3), you reverted the package upgrade. Should I open a new follow-up PR with the same changes as in [this](https://github.com/meditohq/medito-app/pull/705/changes/e1f38fdf6875420f146377331ab5f9144581a2d9) commit?

The old package version doesn't have the changes from this description, nor the [added](https://github.com/meditohq/medito-app/pull/705/changes) `usage` parameter, so now the `develop` branch shouldn't build properly.

PR #886, opened 2026-05-31
fix(ios): deactivate audio session on stop and completion to stop background battery drain

michaelcspeed:
## Problem

iOS users see the app consuming large amounts of background battery — e.g. **1 minute on-screen vs 7h 38m of background activity** in iOS battery settings, with a "used 26% more battery today" warning.

## Root cause

In `lib/providers/player/ios_audio_handler.dart`, `play()` activates the audio session with `session.setActive(true)` under the `.playback` category, but the session was **never deactivated**:

- `pause()` and `stop()` never called `setActive(false)`
- when a meditation reached `ProcessingState.completed`, the session was also left active

With the `.playback` category, an active session keeps the app flagged as "playing audio in the background," so iOS keeps it alive indefinitely after the audio has actually ended — draining the battery for hours.

## Fix

Added a `_deactivateSession()` helper and call it when playback is genuinely done:

1. **On `stop()`** — when the player is closed.
2. **On `ProcessingState.completed`** (when not repeating) — when a meditation finishes naturally.

It uses `AVAudioSessionSetActiveOptions.notifyOthersOnDeactivation` so other audio apps (music, podcasts) can resume once we release the session.

`pause()` is intentionally left active so lock-screen controls and seamless resume are preserved. The background ambient sound is already faded to silence over the final 10s before completion, so releasing the session at completion matches the intended end state.

## Android

Checked — the equivalent bug does **not** exist on Android. There, track completion calls `finishPlayback()` → `stopForeground()` + `stopSelf()`, and `onDestroy()` abandons audio focus and cancels the position-update loop, so the service is torn down when audio ends. (A milder pause-time inefficiency exists on Android — a 1s position-polling loop keeps running while paused — but that's a separate, optional follow-up.)

## Testing notes

- Could not run `flutter analyze` in the dev environment (Flutter not installed); API verified against pinned `audio_session: 0.2.3`.
- Relying on CI test suite + Maestro flows on this PR to validate.

https://claude.ai/code/session_01QVpWUTco6skR9DZVaxaghN

---
_Generated by [Claude Code](https://claude.ai/code/session_01QVpWUTco6skR9DZVaxaghN)_

### DIFF
--- a/android/app/src/main/kotlin/meditofoundation/medito/AudioPlayerService.kt
+++ b/android/app/src/main/kotlin/meditofoundation/medito/AudioPlayerService.kt
@@ -446,8 +446,6 @@ class AudioPlayerService : MediaSessionService(), Player.Listener, MeditoAudioSe
             if (backgroundSoundUri != null) {
                 playBackgroundSound()
             }
-            // Resume the 1s position-polling loop now that playback is active.
-            startPositionUpdates()
         } else if (playbackState == Player.STATE_IDLE ||
             (playbackState == Player.STATE_READY && !primaryPlayer.isPlaying)
         ) {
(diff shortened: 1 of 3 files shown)


michaelcspeed:
Addressed the Copilot review feedback in a36f622:

- **Completion ordering (`ios_audio_handler.dart`)** — the session is now deactivated *before* `_storeTrackCompletion()`, so the network/I/O of stats reporting can't delay releasing the audio session.
- **Dartdoc reference** — changed `[notifyOthersOnDeactivation]` to a plain code span so it no longer tries (and fails) to resolve as a doc link.

Also extended the fix to cover **prolonged pauses** on both platforms:
- iOS: the session is released after a 3-minute grace period if playback stays paused (quick pause/resume stays seamless).
- Android: the 1-second position-polling coroutine is stopped while paused/idle and restarted on resume, so a paused-and-abandoned session no longer wakes the CPU every second.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #900, opened 2026-09-11
Feature/german localization

alexsimon1:
Added german localization


michaelcspeed:
Thanks for this, @alexsimon1, but do you think it will annoy people since the content is still in English?

michaelcspeed:
Also if we release this, would you be able to check that the UI looks OK with German strings? Since they are very often a lot longer than in English. Thanks!

